Issue #51, opened 2023-07-20
Dependency version pinning


tchamberlin:
Currently `requirements.txt` and `requirements_dev.txt` don't have explicit version pins in them. This means that the same commit's tests might execute differently today vs 1 month from now, if e.g. Astropy has released some breaking change

Hatch doesn't provide any dependency management (i.e. a lockfile, like you'd find in Poetry or PDM), so I'd recommend using [pip-tools](https://github.com/jazzband/pip-tools) to manage these. That lets you avoid managing requirements files by hand, but also pins explicit package versions for each commit

mpound:
I try to avoid version pinning except by using >= when required.    It can aggravate users if, e.g. we are pinned to a certain astropy version, but some other package they have is pinned to another.  I suspect most users don't create a venv for every package they use. 

That being said, I have been bitten when trying to manually keep version pinning in sync.  Does pip-tools figure out for instance, that astropy version X requires matplotlib version>Y?

tchamberlin:
This doesn't affect end users at all. For us, the purpose is solely for test reproducibility. For other projects, e.g. web servers, it would allow you to reproduce the known-good environment that all your tests ran against, in production.

The idea is that you have very minimal pins in `pyproject.toml`, as you indicate. But that leaves you open to tests acting differently depending on whether a dependency has released an update or not. So you have some concept of a "lock file". In Poetry that's `poetry.lock`, PDM is `pdm.lock`. Hatch doesn't have a lock file, but we can use `pip-compile` to accomplish basically the same thing.

So, the only people using the requirements files are developers and the automated tests.

PR #52, opened 2023-07-20
Use pip-tools to manage requirements files

tchamberlin:
Implementation of #51

Instructions for use, as per README:
```bash
# Resolve and lock dependencies from `project.dependencies` in `pyproject.toml`
$ pip-compile pyproject.toml --resolver=backtracking -o requirements.txt
# Resolve and lock dependencies from `project.optional-dependencies.dev` in `pyproject.toml`
$ pip-compile pyproject.toml --extra dev --resolver=backtracking -o requirements_dev.txt
```

Note that in repos where I use pip-tools, I typically do so via a pre-commit hook which runs whenever `pyproject.toml` is modified. That way I never have to think about this


tchamberlin:
This fails due to [Astropy 5.3 dropping support for Python 3.8](https://docs.astropy.org/en/stable/changelog.html#other-changes-and-additions). I guess there are a few options:

1. Specify different versions of Astropy depending on Python version
2. Pin the max Astropy version to <5.3 (seems bad)
3. Drop support for Python 3.8, as Astropy has now done
4. Don't pin astropy version to an exact version

I'll take a pass at (1)

tchamberlin:
Okay, this looks happy now, although I'd like to see if I can clean up a bit more, since there are now multiple versions of the requirements files in there.

But it does work -- there is a set of requirements for 3.8, which is limited to `numpy<1.25` and `astropy<5.3`, and another set for 3.9+, which does not care which numpy/astropy you have

PR #72, opened 2023-08-23
Fix syntax error in release.yaml

tchamberlin:
"Build with hatch" step was missing a `run:` block


mpound:
glad it was something simple!

PR #74, opened 2023-08-23
Run pre-commit on all files

tchamberlin:
- Update all pre-commit hooks
- Run pre-commit hooks on all files
- Add the "pre-commit" commit to `.git-blame-ignore-revs`, as per https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

Note: this **will** cause merge conflicts if you branched from `main` prior to this PR. To resolve, I _think_ you should be able to:

1. Checkout `.pre-commit-config.yaml` from `main`
2. Run `pre-commit install` to install the hooks
3. Run `pre-commit run --all-files` to run hooks on all files (which was already done in `main`, but you will need to do in order to "sync" your branch with `main`)
4. Try the merge again


tchamberlin:
Okay, I finally have this mostly sorted out I think. Here's a summary of my changes:

### `pyproject.toml`

I did some cleanup here:
- Bumped minimum Python version to 3.9 (also fixed all references to 3.8 I could find)
- Add `dev` block to `optional-dependencies`. This defines the requirements to run tests, and to build docs
- Bumped Development Status classifier to "Alpha"

### Hatch

- Add `test` environment. This basically allows you do e.g. `$ hatch run test:cov` to run all unit tests across all supported versions of Python, with coverage
- Add new `scripts` for easy linting/testing/doc-building. I think these should replace any `Makefile` or shell-script approaches. For example:
  - `hatch run lint` will run the linter (Ruff)
  - `hatch run docs` will run the docs in auto-build mode
  - `hatch run test` will run unit tests
  - `hatch run test:matrix` will run unit tests across all supported Python versions (similar to what our GitHub action does, but local)

### Ruff

I have replaced PyLint with [Ruff](https://beta.ruff.rs/docs) in all of my projects. It is _much_ faster, and replaces several tools. For example, if you run `hatch run lint --fix`, it will:
- Lint all files (see `tool.ruff.select` in `pyproject.toml` for details of what it's checking by default)
- Organize imports

### Black

[Black](https://black.readthedocs.io/en/stable/) has been run on all files. Things you'll probably notice:
- It does _not_ wrap long comment lines, but everything else has been wrapped to 120 characters
- All quotes are double quotes, except in situations where double quotes are nested within single
- The number of lines between functions/classes/imports has been normalized
- Long argument lists have been wrapped and take up much more vertical space

### pre-commit

Core principle: fix code issues **before** they make it into the commit history.

Manually running linters/formatters, fixing typos, wrapping lines is all quite tedious. Also, if only some developers are using Black, then there will inevitably be very noisy diffs -- exactly the thing it is trying to prevent.

To avoid the tedium, we're using pre-commit. It should be as simple as:

```sh
pipx install pre-commit  # you can also just install it directly into your venv with pip, but this is fragile
# Install the git hooks
pre-commit install
```

Note that you will need to have `python3.9` in your `PATH`, or you will get errors.

Now, whenever you try to commit code, you will see something like this:

```sh
$ git commit ...
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check for added large files..............................................Passed
debug statements (python)............................(no files to check)Skipped
mixed line ending........................................................Passed
ruff....................................................................Passed
black................................................(no files to check)Skipped
pip-compile..........................................(no files to check)Skipped
```

In this situation, no checks failed, so the commit succeeds. You will often see failures here, which will prevent the commit from happening. Most of the time, re-staging the files and re-attempting the commit will work -- errors are typically fixed as they come up. 

After you push, the pre-commit hooks run again as a GitHub Action (e.g. https://github.com/GreenBankObservatory/dysh/actions/runs/5979685062/job/16224280168). This prevents issues where someone skips checks or never installs pre-commit at all.

mpound (commented):
looks good.

PR #84, opened 2023-10-05
Cat devel merger

vcatlett-nrao:
Adding documentation changes, testing changes, and misc file changes (README, .gitignore, requirements.txt)


.github/workflows/python-package.yml
vcatlett-nrao, line 1: Renamed this to build.py since build.py was very similar, just with a hatch build on top and no pytest run

.github/workflows/release.yml
vcatlett-nrao, line 1: This test was consistently failing for every push (even though it wasn't trying to run. It was just failing a GitHub syntax check). I fixed the yml syntax so it would stop failing.

vcatlett-nrao:
Makefile is missing for the docs

PR #85, opened 2023-10-05
Cat devel merger

vcatlett-nrao:
Hopefully I don't have to close this again! This is all the documentation and test changes. I'll go through and leave comments on changes that might be confusing.


.env.template
vcatlett-nrao, line 1: Moved to top level of directory so there's not a separate docs-only environment for developers. Can add more environment variables for development here as needed.

.github/workflows/pyinstaller.yml
vcatlett-nrao, line 1: Test to check if the pyinstaller run fails for the GUI. GUI code not added yet, but I'm adding this with the tests so it runs when I do add the GUI code.

.github/workflows/python-package.yml
vcatlett-nrao, line 1: Combined with build.yml (see above)

.github/workflows/release.yml
vcatlett-nrao, line 1: This was failing on every push because of a YAML syntax error even though it wouldn't actually try to run for said branches. Fixed that error so it's stops making so many workflow fails.

.gitignore
vcatlett-nrao, line 1: I keep accidentally pushing stuff I shouldn't. This really keeps me from doing that.

README.md
vcatlett-nrao, line 1: Fixed a typo and added instructions for specifying which Python hatch should use (since the system python at GBO is way out of date)

docs/requirements.txt
vcatlett-nrao, line 1: Combined with the top-level requirements

docs/.env.template
vcatlett-nrao, line 1: Moved to top level of repo

docs/source/examples/positionswitch.rst
vcatlett-nrao, line 1: For a lot of these, I'd moved the image directories to be paired with their folders, not all together in _static/. And I added alt text to images.

docs/source/for_developers/doc_standards.rst
vcatlett-nrao, line 1: Updated doc build instructions based on the changes to the requirements.txt and location of .env.template

requirements.txt
vcatlett-nrao, line 1: Should be the same requirements as before (check the dependency diffs), just rearranged from when I was trying to make sense of my environment issues.

src/dysh/plot/tests/test_specplot.py
vcatlett-nrao, line 1: I had to remove the pathlib dependency in order for pyinstaller to work for the GUI (it wouldn't even run if it found pathlib somewhere). Commenting out the dysh_root line doesn't seem to affect this test, and we can use os anywhere else we need to use paths.

src/dysh/spectra/tests/test_scan.py
vcatlett-nrao, line 1: Commenting out unit test that always fails. We need to set an actual threshold for this test that's not 0.

requirements.txt
vcatlett-nrao: Edits to requirements.txt here were adding ones identified as missing but really were still installed as dependencies of others. But I put them back in there to be explicit.

docs/source/design/structure.rst
vcatlett-nrao, line 1: Messing around with adding some structure diagrams

docs/source/for_developers/terminal_messages.rst
vcatlett-nrao, line 1: Info about the rich text printing I added

.github/workflows/build.yml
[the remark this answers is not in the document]
mpound: I had build.yml separate because it is used when making releases ( "on: release:").   This commit changes that function entirely, so I can't accept this change.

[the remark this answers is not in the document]
mpound: The recommended way of doing this is to invoke build action from whatever other yml you want to do a build.   We should be moving towards smallish action ymls that can be invoked in different places.

.github/workflows/release.yml
mpound: for a release action we don't want to build on all python versions just the latest which is why it was 3.x

.github/workflows/build.yml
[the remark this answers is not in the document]
mpound: also the workflow_dispatch is required so we can run this action manually from github when needed

.github/workflows/release.yml
vcatlett-nrao: Reverted python version to 3.x. Syntax error stays resolved.

.github/workflows/build.yml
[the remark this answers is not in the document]
vcatlett-nrao: Reverted build.yml to its original version and renamed the other test build-and-pytest.yml

docs/source/design/structure.rst
mpound: FYI Obsblock is being removed.  ScanBlock will replace.  So we have to remember to update this diagram :-)

mpound (approved):
Looks good. Thanks for reverting those YML files.

Issue #90, opened 2023-10-16
Makefile broken for new docs directory structure


mpound:
make fails in src/dysh/docs, e.g.

make html
Running Sphinx v6.1.3

Configuration error:
config directory doesn't contain a conf.py file (/home/mpound/src/dysh/docs)
make: *** [Makefile:20: html] Error 2

This is because conf.py was move to the source directory.   
There are two possible solutions
1) move Makefile to source.  However this pollutes the source directory with _build. 
2) move conf.py to docs.   Then it needs to be modified so that e.g. make html works.  

As it is I don't see how readthedocs for version 0.2.0 was succeeding.

vcatlett:
See https://github.com/GreenBankObservatory/dysh/pull/96

PR #95, opened 2023-10-18
Fix issue #90

vcatlett:
Changes to fix the bug reported in issue https://github.com/GreenBankObservatory/dysh/issues/90. It was mainly just fixing paths in the .readthedocs.yaml file. Requirements.txt file also had some dependency issues for Python 3.8 that I resolved. Docs now build with the same hatch environment. See https://dysh.readthedocs.io/en/cat-devel/for_developers/doc_standards.html for the new docs instructions.


vcatlett:
Sorry, wrong branch

PR #96, opened 2023-10-18
clean up #90 bug fix

vcatlett:
Changes to fix the bug reported in issue https://github.com/GreenBankObservatory/dysh/issues/90. It was mainly just fixing paths in the .readthedocs.yaml file. Requirements.txt file also had some dependency issues for Python 3.8 that I resolved. Docs now build with the same hatch environment. See https://dysh.readthedocs.io/en/cat-devel/for_developers/doc_standards.html for the new docs instructions.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,17 +7,17 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
-  os: ubuntu-22.04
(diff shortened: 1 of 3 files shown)


astrofle (approved):
It all looks good to me, and the build in ReadTheDocs is successful. Thanks Cat!

Issue #106, opened 2023-11-15
Create basic Dysh shell


tchamberlin:
```sh
$ dysh
```

Should launch the user into an IPython session where relevant dysh modules have been pre-loaded.

Optional: Dysh-specific IPython profile is used, so that there's no leakage between dysh and non-dysh IPython sessions

teuben:
Here's the example I cooked up, with the echo's just an example:

#
if [ -e ~/.ipython/profile_dysh ]; then
    ipython --profile=dysh
else
    echo "Generating a new ipython dysh profile"
    ipython profile create dysh
    # or copy from a template
    f=~/.ipython/profile_dysh/startup/90-numpy.py
    echo 'import numpy as np'
    echo 'import numpy.ma as ma'
    echo 'import math'
    echo 'import matplotlib.pyplot as plt'
    echo 'import pandas as pd'
    echo 'import dysh'
    echo 'print("dysh is ready")'
fi

tchamberlin:
My implementation does most of the above -- uses a `dysh` profile by default, which will be automatically created for the user whenever they first launch `dysh`. It imports a few sensible-seeming things and alerts the user of what it has done

This definitely still needs some work, but I'm calling this done for now. We will need to review the draft together before merging anything in, though

PR #108, opened 2023-11-15
Create dysh entry point/interactive shell

tchamberlin:
Implementation of #106 

- Create a `$ dysh` entry point
- `$ dysh` will load an IPython shell, and preload several convenvience functions
- `$ dysh /path/to/sdfits1.fits /path/to/sdfits2.fits` will pre-load all given SDFITS paths into `sdfits_files` variable


tchamberlin:
This is essentially a thin wrapper around `IPython.start_ipython`. `dysh` directly handles a few options, but passes the rest of `argv` through to `IPython.start_ipython`. This means that users can use whatever flags they are used to for `ipython`.

I'm basing this conceptually off of `django-extensions` `shell_plus`: https://django-extensions.readthedocs.io/en/latest/shell_plus.html . It would probably be good to actually look at the code and see how they handle support for multiple shells, and the configurations thereof.

Here is the current base behavior:

```sh
$ dysh
Python 3.11.0 (main, Nov 22 2022, 12:03:52) [GCC 12.2.1 20220819 (Red Hat 12.2.1-2)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: dysh

--------------------------------------------------------------------------
                         Welcome to Dysh v0.2.0

    Example usage: https://dysh.readthedocs.io/en/latest/example.html
    Bug repots:    https://github.com/GreenBankObservatory/dysh/issues

    For help with a Dysh routine from the command line,
    use the builtin 'help'. e.g.:

        help(GBTFITSLoad)

    The following items have been made globally available for convenience:
        pd (from pandas)
        np (from numpy)
        GBTFITSLoad (from GBTFITSLoad)
        Table (from Table)
        fits (from astropy.io.fits)

   To suppress this banner, start with --no-banner flag
--------------------------------------------------------------------------

In [1]: 
```

mpound:
hatch is not getting installed, so the builds are failing.   somewhere pip install requirements_dev.txt has been removed??  it is in build.yml so I'm not sure what is happening.   it is not in hatch_and_pytest.yml, which is incorrect, but that workflow is not being invoked.

mpound:
Thomas - can you submit a new PR for this?  release-0.2.0 just got a significant update.
thanks
Marc

PR #110, opened 2023-11-16
Multifits

mpound:
Clean merge of multifits into release 0.2.0


tchamberlin:
There are two issues with pre-commit here:
1. Neither `multifits` nor `release-0.2.0` are compliant with their own pre-commit hooks. `pre-commit run --all-files` should be a no-op, but it results in changes on both of these branches
2. `.pre-commit-config.yaml` is not identical between these two branches

A potential path forward:

1. Create `release-0.2.0-tmp`, off of current `release-0.2.0`
2. On `release-0.2.0-tmp`, check out `multifits` version of `.pre-commit-config.yaml`
3. On `release-0.2.0-tmp`, do `pre-commit run --all-files`, then commit something like `"Run pre-commit on all files"`
4. On `multifits`, do `pre-commit run --all-files`, then commit something like `"Run pre-commit on all files"`
5. _Then_ start the merge

That way you have the same formatting applied on both branches, which _should_ remove any formatting-specific diffs/conflicts

Note that this will only be an issue short term, because we still don't have the pre-commit hooks run/enforced on all branches

mpound:
Thanks, I will do this.

PR #115, opened 2023-11-20
Release resolve conflicts

mpound:
All changes go one way -- from release_resolve_conflicts to release-0.2.0.  There is nothing added to release-0.2.0 that is not already in release_resolve_conflicts,


astrofle:
It seems that in windows the difference between `GBTIDL` and `dysh` is larger than in MacOS or Ubuntu.
You could change [this line ](https://github.com/GreenBankObservatory/dysh/blob/d40bf4dca3247679ffcbcd9fbde234f58ce71a51/src/dysh/spectra/tests/test_scan.py#L125) to be < 5e-16 and it should pass.
We should still understand why this is the case.

PR #118, opened 2023-11-20
First draft of dysh shell

tchamberlin:
New version of #108 

Implementation of issue #106 

- Create a `$ dysh` entry point
- `$ dysh` will load an IPython shell, and preload several convenvience functions
- `$ dysh /path/to/sdfits1.fits /path/to/sdfits2.fits` will pre-load all given SDFITS paths into `sdfits_files` variable


mpound (approved):
line 17 of shell.py has a typo.  other than that ok.

PR #119, opened 2023-11-20
Fix invalid escape sequences in docstrings

tchamberlin:
These were resulting in SyntaxWarning in Python 3.12:

```plain
dysh/src/dysh/plot/specplot.py:10: SyntaxWarning: invalid escape sequence '\*'
  """
dysh/src/dysh/plot/specplot.py:91: SyntaxWarning: invalid escape sequence '\*'
  """
dysh/src/dysh/plot/specplot.py:167: SyntaxWarning: invalid escape sequence '\*'
  """Set x and y labels according to spectral units
dysh/src/dysh/spectra/scan.py:155: SyntaxWarning: invalid escape sequence '\d'
  """Average all polarizations in all scans in this ScanBlock
dysh/src/dysh/spectra/scan.py:175: SyntaxWarning: invalid escape sequence '\d'
  """Average all times and polarizations in all scans this ScanBlock
dysh/src/dysh/spectra/scan.py:697: SyntaxWarning: invalid escape sequence '\d'
  """
```

### DIFF
--- a/src/dysh/plot/specplot.py
+++ b/src/dysh/plot/specplot.py
@@ -10,7 +10,7 @@ class SpectrumPlot:
     """
     The SpectrumPlot class is for simple plotting of a ~spectrum.Spectrum
     using matplotlib functions.   Plots attributes are modified using keywords
-    (\*\*kwargs) described below SpectrumPlot will attempt to make smart default
+    (\\*\\*kwargs) described below SpectrumPlot will attempt to make smart default
     choices for the plot if no additional keywords are given.
     The attributes are "sticky" meaning that an attribute set via
     instantiation or by the `plot()` method will stay set until changed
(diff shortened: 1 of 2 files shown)


mpound:
Yeah, these were to have proper **kwargs shown in documentation.  I'll accept this change and we can see what it does to docs.

mpound (changes requested):
The double \ in the :math: lines is incorrect. These are latex symbols with a single \

tchamberlin:
Ah okay! I've reverted all of this and changed to raw-string-literal docstrings. These are already being used elsewhere, and the numpy examples use it as well: https://github.com/numpy/numpydoc/blob/main/doc/example.py

mpound (approved):
Also, FYI the r at the beginning of some docstrings allows one to use latex math markup in the documentation itself.

PR #123, opened 2023-11-21
Make ipython a core dependency

tchamberlin:
`$ dysh` (`dysh.shell`) requires `ipython`, so it needs to be migrated from an optional to a "core" dependency


tchamberlin:
I can add a unit test for this

tchamberlin:
Please don't accept draft PRs; once they are done they'll be marked as ready

PR #131, opened 2023-11-21
Update docs for beta

astrofle:
This PR updates the examples to work with ScanBlocks. It is mainly changing from `ta.plot()` to `ta[0].plot()`.
It also changes the `GBTSDFITSLoader.summary` to print when using `show_index=False`.

### DIFF
--- a/docs/source/examples/positionswitch.rst
+++ b/docs/source/examples/positionswitch.rst
@@ -145,7 +145,7 @@ The system temperature array (`numpy.ndarray`) is stored in `tsys`
 
 .. code:: python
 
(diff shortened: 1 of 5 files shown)


mpound (changes requested):
If we are changing gbtfitsload.py to print() the index if show_index=False, then we should default to show_index=True so that the user does not have to type it to get pretty output in a notebook.   i.e., summary() with no arguments should to the prettiest thing.

astrofle:
Agree. Changed now.

mpound (approved):

Issue #132, opened 2023-11-22
gettp fails when loading multiple SDFITS files


astrofle:
The following code fails using `release-0.2.0rc`

``` Python
from dysh.fits.gbtfitsload import GBTFITSLoad
import dysh.util as util

testdata = util.get_project_testdata()
data_path = testdata  / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas"

sdf = GBTFITSLoad(data_path, verbose=False)

tp_scans = sdf.gettp(6, ifnum=2, plnum=0)
```

The error is:
``` Python
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[9], line 1
----> 1 tp_scans = sdf.gettp(6, ifnum=2, plnum=0)

File /home/sandboxes/psalas/Dysh/dysh/src/dysh/fits/gbtfitsload.py:508, in GBTFITSLoad.gettp(self, scan, sig, cal, bintable, **kwargs)
    505         print("CALROWS on len=", len(calrows["ON"]))
    506         print("fitsindex=", i)
--> 508     g = TPScan(
    509         self._sdf[i], scan, sigstate[sig], calstate[cal], tprows, calrows, bintable, kwargs_opts["calibrate"]
    510     )
    511     scanblock.append(g)
    512 if len(scanblock) == 0:

File /home/sandboxes/psalas/Dysh/dysh/src/dysh/spectra/scan.py:228, in TPScan.__init__(self, gbtfits, scan, sigstate, calstate, scanrows, calrows, bintable, calibrate)
    225 # print("BINTABLE = ", bintable)
    226 # @TODO deal with data that crosses bintables
    227 if bintable is None:
--> 228     self._bintable_index = self._sdfits._find_bintable_and_row(self._scanrows[0])[0]
    229 else:
    230     self._bintable_index = bintable

IndexError: list index out of range
```

PR #134, opened 2023-11-22
Patches for `gettp`

astrofle:
This patch fixes Issue #132.
Now `gettp` will skip the creation of a `TPScan` if the input selection results in zero rows.


mpound (approved):
Thanks.  I had fixed a similar issue in subbeammod.

PR #143, opened 2023-12-01
new core file, veldef function and test

mpound:
This is the beginning of the doppler work.   Add a core.py to dysh.fits and implement function to decode VELDEF string. Also test code for that function.


mpound:
missing a file I think.

PR #145, opened 2023-12-04
Veldef function

mpound:
one more time....utility code for velocity conventions.  some test files renamed due to pytest complaining that files have the same name.


astrofle (approved):
Looks good.

PR #147, opened 2023-12-04
Fix for getspec

astrofle:
`getspec` was not working after the changes made to support multiple fits files.
We are using this function to test the baseline removal.

I'm setting the PR to go to `main`, but I'm not sure that is what we want.


mpound (approved):
ultimately the test should check the the correct spectrum was fetched.  This could perhaps be done by comparing header information or data to what is expected.   I'll approve this but please submit an issue to upgrade this test.

PR #150, opened 2023-12-05
Various minor tweaks to clean up the repo

tchamberlin:
As per #140, various unused files have been removed (GUI code, attic, etc.). Note that `wget` has _not_ been removed in this PR; this is functional change that I'll address separately

As per #139, the dependencies in `pyproject.toml` have been updated/reorganized

PR #151, opened 2023-12-05
Updates to GitHub Actions/Workflows

tchamberlin:
- Enable caching of pip reqs
- Enable code coverage generation and reporting
- Removed unused workflow files
- Update README to add badges
- Regenerate requirements.txt (as a lockfile)
- Remove `requirements_dev.txt`; not used
- Rename `hatch-and-pytest` to just `ci`

Wait to merge this until #150 is accepted


mpound (approved):
Reason for removing check-docstring-first?  Not something we care to enforce?

tchamberlin:
Well I had thought it was redundant, as docstrings are checked by [`ruff`](https://docs.astral.sh/ruff/rules/#pydocstyle-d) (via their re-implemntation of pydocstyle). But:
1. Docstrings aren't being checked in our config (too noisy)
2. The error that `ruff`/`pydocstyle` produces (`D100`) is much less useful: `"Missing docstring in public module"`

So, I'll put it back. And at some point hopefully we can enable the `pydocstyle` checks (`D`)

PR #152, opened 2023-12-05
Remove wget as dependency; replace with Astropy's download_file

tchamberlin:
I've spun this off into a new PR, because it introduces some functional changes.

Previous behavior:

```py
import wget

filename = "TGBT21A_501_11.raw.vegas.fits"
if not os.path.isfile(filename):
    url = f"http://www.gb.nrao.edu/dysh/example_data/onoff-L/data/{filename}"
    print(f"Downloading {filename}")
    wget(url,out=filename)
    print(f"\nRetrieved {filename}")
else:
    print(f"{filename} already downloaded")
```

Given input path `filename`, this uses `wget` to download` the data to the current working directory, keeping the original filename

In contrast, the new behavior:

```py
from dysh.util.data import download_file

filename = "TGBT21A_501_11.raw.vegas.fits"
url = f"http://www.gb.nrao.edu/dysh/example_data/onoff-L/data/{filename}"
path = download_file(url)
print(f"Retrieved {filename}")
```

Output:

```sh
$HOME/.dysh/cache/download/url/9bf83516cba5e59a687fcb5d1e5c24c9/contents
```

Astropy has stuck the downloaded file into the Dysh cache (located by default at `$HOME/.dysh`, but can be changed via Astropy config). Subsequent downloads of the same URL will automatically retrieve the file from the cache instead of downloading it again. Users can force-update the cached version via `download_file(url, cache="update")`. See https://docs.astropy.org/en/stable/api/astropy.utils.data.download_file.html#download-file for more details

So, the core issue here is that, without `wget`, we don't have a way of actually downloading files to arbitrary locations. I'm not sure that we have a use case for that, but if we do this probably isn't a great solution for that. Astropy's `download_file` will download to `/tmp` by default, so we could download then move to a user-specified location. But this is going to result in a copy if the source and destination filesystems differ


astrofle:
The only potential issue I have with using `astropy`'s solution is that the "filename" does not say anything about what the data is. I think this will require some explanation on our end -- which is doable.

teuben:
To note that the get_dysh() method would be impacted by this implementation. Developers need to be educated about the two worflows or they will wind up with duplicate large files

Issue #153, opened 2023-12-05
Subbeamnod routine cannot access plnum


jfrothingham:
Unable to retrieve a subbeamnod scan because of apparent problems accessing 'plnum'

I've been following the SubBeamNod example (https://dysh.readthedocs.io/en/latest/examples/subbeamnod.html) with a different dataset.

I am using dysh-0.2.0b with Python3.11 on a computer that I think is running RedHat 8. (Thales data reduction machine)
Data is located at /home/scratch/jfrothin/testdata/AGBT19B_096_08.raw.vegas in the Green Bank filesystem.

Here is a screenshot of the error I'm seeing. I get this error regardless of whether I specify kwargs or not, and regardless of which scan I use.

![image](https://github.com/GreenBankObservatory/dysh/assets/16109523/d00c6584-9503-4042-84e8-46fe638556cf)

However, sdfits.summary() shows that this scan does have values for PLNUM and FDNUM.

![image](https://github.com/GreenBankObservatory/dysh/assets/16109523/4d442a25-8714-4f30-8744-97926e50c50f)

tchamberlin:
We will need to reproduce the error somehow -- is the dataset you are using something that we could test against?

jfrothingham:
Yes, I think you can test against it. It's a dataset we were considering using as a training example

tchamberlin:
I can reproduce this:

```py
sdfits = GBTFITSLoad("/home/scratch/jfrothin/testdata/AGBT19B_096_08.raw.vegas")
sbn = sdfits.subbeamnod(scan=18, fdnum=0, plnum=1)
```

Indeed, `plnum` is not defined, since `rx` is not `"Rcvr26_40"`. But upon fixing that, I get

```
File <snip>/dysh/spectra/scan.py:228, in TPScan.__init__(self, gbtfits, scan, sigstate, calstate, scanrows, calrows, bintable, calibrate)
    225 # print("BINTABLE = ", bintable)
    226 # @TODO deal with data that crosses bintables
    227 if bintable is None:
--> 228     self._bintable_index = self._sdfits._find_bintable_and_row(self._scanrows[0])[0]
    229 else:
    230     self._bintable_index = bintable

IndexError: index 0 is out of bounds for axis 0 with size 0
```

tchamberlin:
[...]

PR #156, opened 2023-12-06
Provide default value for plnum

tchamberlin:
As per #153, this simply adds (what I hope is) a sensible default value for `plnum`


mpound:
As we know, plnum can depend on fdnum so this code will need checks for consistency of inputs.  Also should probably have a default kwargs_opts["plnum"] = 1 as with other params.   I will approve this for now to fix the immediate bug, and revisit the code more comprehesively later.

Issue #93, opened 2023-10-18
Spectrum.plot() remembers old defaults?


teuben:
from a tutorial example, i tried

     ta.plot(xaxis_unit="km/s",yaxis_unit="mK",ymin=-100,ymax=500,xmin=3000,xmax=4500)

being unhappy what I saw (that's a different issue) I wanted it auto-scaled, so naively tried

     ta.plot(xaxis_unit="km/s",yaxis_unit="mK")

but was surprised that this didn''t work. I tried the pythonesq

      ta.plot(xaxis_unit="km/s",yaxis_unit="mK",ymin=None,ymax=None,xmin=None,xmax=None)

and that worked. But I think this behavior is not what we should expect. Is the "remembering of old default" something that's creeping into dysh , and do we really want this?  [i think i don't]

teuben:
the documentation in SpectrumPlot() does say the arguments are sticky. So in this sense my issue isn't an issue.

However, this is not the python norm. I am not convince we need to follow this convention, or did that train already depart?  It smells too much like the tget/tput in aips/casa 'go' envirnoment.   I thought we would follow python with least surprises.

astrofle:
We could ask the beta testers about this. I'll make sure to include an example and question about the sticky behavior.

teuben:
It seems the title= is still remembered from one to the next plot

astrofle:
It seems that any attribute that is set the first time is remembered.
For example:
```Python
from dysh.spectra import Spectrum
s = Spectrum.fake_spectrum()
s.plot(ymin=0.5)
s.plot() # Will produce the same plot as above, even though there's data going up to y~1
```
I kind of like sticky attributes though (and GBTIDL certainly is sticky).

PR #164, opened 2023-12-11
remove stickiness of plot attributes

mpound:
This should solve issue #93.   Plot stickiness is removed.


vcatlett:
Looks good to me. We should eventually add a way to update the defaults for all plots, but starting with behavior that's limited to individual plots is good to remove the unexpected stickiness described in Issue 93.

mpound:
Yes, we have talked about an rc file for plot defaults.  That can be a general file for all plots.

vcatlett (approved):
Looks good to me

Issue #148, opened 2023-12-04
Azimuth values in GBTFITSLoad.summary()


dbautista98:
When printing the observation summary with `GBTFITSLoad.summary()` the azimuth values for an observation are different in GBTIDL and dysh. In GBTIDL the Az is 352.2 and 187.2 but with dysh, it is 262.88 and 268.9

Here is the output with GBTIDL 
![gbtidl](https://github.com/GreenBankObservatory/dysh/assets/60448615/3cd72b76-f150-4737-8713-e696c0cc21f8)

Here is the output with dysh
<img width="1008" alt="dysh" src="https://github.com/GreenBankObservatory/dysh/assets/60448615/9f04e45c-589a-4f51-b2db-901992bbeff0">

astrofle:
Hi Daniel,
Thanks for opening up this issue.
`dysh` takes the mean of the azimuth and elevation when producing the summary, while `GBTIDL` reports the first value for the scan. 
We will document this better so it is clear. Would that be a proper solution from your perspective?

dbautista98:
Got it, thanks for the clarification! Yes that sounds like a good solution.

mpound:
To clarify, the average is taken for columns VELOCITY, RESTFREQ, DOPFREQ, AZIMUTH, ELEVATIO.

PR #167, opened 2023-12-14
Update docs and add test files

astrofle:
Updated docs and added test files for Doppler corrections.

The documentation updates are in response to the beta testing feedback.
See Issue #148


mpound:
For each scan in the summary() row, dysh takes the mean of numeric columns
VELOCITY, RESTFREQ, DOPFREQ, AZIMUTH, ELEVATIO
so these may differ from GBTIDL if it is reporting only the first integration values. (If verbose=True you get all the integrations)

Please update the note.

mpound (changes requested):
See conversation about other numeric columns being averaged.  Docs should be tweaked.

mpound (approved):

PR #178, opened 2024-01-05
Doppler frame conversions. 

mpound:
This PR contains many updates.  In getting doppler to work there was lots of ancillary code, either new code or fixing up old code.

1.  The implementation to switch between doppler frames:
    -  the code itself
    -  test code test_doppler_frames.py
    -  SDFITS test file
 2.  Spectrum.make_spectrum - factory method for making spectra, now used by all Scan classes, which makes them cleaner
 3.  Bug fixes in *FITSLoad [e.g. write_scans()]
 4. new Observatory class wrapper for astropy.EarthLocation, with special case for GBT location since GBO's location is more accurate than astropy's
 5. new coordinates module, required to keep clean separate between spectra and fits modules. 
 
Notes:
- The doppler frame test checks that the frequency difference is less than 2Hz in most cases.  Astropy defines Galactocentric and LSRD coordinates differently than GBTIDL, so they are allowed more leeway.  
-  I have left off TOPO from the doppler test because there is a bit more code to write to make that work, but I thought it more important to get these changes in because there are so many.  
- Ideally we should test these functions for a few sources in different parts of the sky since the velocity vectors will change.  Add to wish list.
- There is some code that I wrote, mainly in coordinates module, that is not used in the frame conversion per se, but I needed it during development, I'm leaving it in for now.  It is not guaranteed to work all the time.


mpound:
Specifically, for LSRD, the difference is ~25 Hz.   This is because GBTIDL uses 16.6 km/s towards RA=17:49:58.7 hours, DEC=28.07.0 J2000, whereas astropy uses  U=9 km/s, V=12 km/s, W=7 km/s in Galactic coordinates or 16.552945 km/s  towards l=53.13 b=25.02.    Our test source is a RA=9h,DEC=40. This is a good reason to test in different parts of the sky to see how large this difference really gets.

astrofle (approved):
The new functionality works well.
There are a few items that need to be addressed, but we will do that in a separate branch.

PR #181, opened 2024-01-08
153 subbeamnod routine cannot access plnum

mpound:
This fixes issue #153 as well as the associated test.


astrofle:
This never made it to the main branch

PR #193, opened 2024-01-19
189 display velocity frames correctly in specplot

mpound:
This fixes inconsistency in the spectral axis when plotting with alternate velocity frames or doppler conventions.   Also some cleanup of lower level code for frame/convention switching. 

The notebook developer/test_specplot.ipynb is a simple visual test.


teuben:
TGBT21A_501_11.raw.vegas.fits is needed in the test, the current test notebook has an absolute path, isn't there a new method to derive correctly for anybody's dysh environment?

mpound:
I will revise the notebook to do something intelligent with data path (despite what I said in email!)

mpound:
I've pushed a new notebook with regression sanity check and removed a debug print statement.

teuben (approved):
looks good now

PR #209, opened 2024-03-13
Release 0.2.0

mpound:
This merges in the missing items from release-0.2.0 including subbeamnod-153.


astrofle:
This would change release GitHub action from running on the matrix of OSs to `ubuntu-latest`. Is that correct?

mpound:
This is only for the release action which uploads the results of hatch build to PyPI.   Everything should have been tested prior on $matrix.os via PRs if we are doing things right.

astrofle (approved):

Issue #211, opened 2024-03-14
Wrong number of integrations being reported


astrofle:
When a spectral window or feed is out of sync, `summary` reports the wrong number of integrations.

To reproduce this (on GBO):
```Python
from pathlib import Path

import pandas as pd
from dysh.fits import GBTFITSLoad
from pandas.testing import assert_series_equal

def read_gbtidl_summary(filename, idx=1):
    """
    """
    with open(filename, "r") as log:
        lines = log.readlines()
    lines.pop(idx)
    tmp = Path(f"{filename}.tmp")
    with open(tmp, "w") as f:
        for line in lines:
            f.write(line)
    
    df = pd.read_fwf(f"{filename}.tmp")
    # Clean up.
    tmp.unlink()
    
    return df

cols = {"SCAN": "Scan",
        "OBJECT": "Source",
        "VELOCITY": "Vel",
        # "PROC": "Proc", # GBTIDL trims the names.
        "PROCSEQN": "Seq",
        "# IF": "nIF",
        "# INT": "nInt",
        "# FEED": "nFd"
       }

path = "/home/dysh/acceptance_testing/data"
projs = [#"AGBT18B_354_03", 
         #"AGBT20B_336_01", 
         #"AGBT22A_325_15", 
         "AGBT19A_080_01",
         "AGBT17B_004_14",
         "AGBT16B_392_01",
         "AGBT15B_228_08",
         "AGBT18A_503_02",
         "AGBT19A_473_41",
         "AGBT13A_240_03",
         "AGBT14B_480_06",
         "AGBT19B_096_08"
        ]

for proj in projs:
    try:
        sdf = GBTFITSLoad(f"{path}/{proj}/{proj}.raw.vegas/")
    except:
        sdf = GBTFITSLoad(f"{path}/{proj}/")
    dysh_df = sdf.summary()
    gbtidl_summary = read_gbtidl_summary(f"{path}/{proj}/gbtidl/{proj}.summary")
    print(proj)
    for col in cols.items():
        assert_series_equal(dysh_df[col[0]].sort_values(), 
                            gbtidl_summary[col[1]].sort_values(), 
                            check_dtype=False, 
                            check_names=False,
                            check_index=False,
                           )
```

PR #212, opened 2024-03-14
Fix bug reporting wrong number of integrations

astrofle:
This fixes Issue #211 
Not elegant at all, but I'm dying to get summary marked as done.


astrofle:
In `GBTIDL` parsing the number of integrations from the time stamps works fine, because `GBTIDL` index files are for each sampler (i.e. VEGAS bank), whereas `dysh` merges all of them into a single `DataFrame`. So, if one of the samplers is out of sync, the time stamps will be different, and when we count all of them together it will look like there are more integrations.

mpound:
Please enhance the comment "Take care of out-of-sync samplers" to "For counting integrations, take care of out-of-sync samplers by just looking at the first instance of FEED, PLNUM, and IFNUM".
Other than that good to go!

astrofle:
Updated comment.

mpound (approved):

PR #214, opened 2024-03-15
208 spectrum arithmetic

astrofle:
This tries to fix the broken `Spectrum` arithmetic after including the code to handle Doppler definitions.

This is not yet complete, in the sense that it does not include all the operations, nor all the ways we should handle the metadata.


src/dysh/spectra/spectrum.py
```diff
@@ -572,11 +572,11 @@ def __add__(self, other):
...
     def __sub__(self, other):
-        subtraction = self.sub(other, **{"handle_meta": self._add_meta})
+        subtraction = self.subtract(other, **{"handle_meta": self._add_meta})
         subtraction._target = self._target
         return subtraction
 
-    def add_meta(self, operand, operand2=None, **kwargs):
+    def _add_meta(self, operand, operand2=None, **kwargs):
         meta = deepcopy(operand)
         meta["EXPOSURE"] = operand["EXPOSURE"] + operand2["EXPOSURE"]
```
mpound, line 581: Why does operand2 default to None instead of uninitialized?   If it remains None, line 581 will raise an exception.

testdata/AGBT05B_047_01/gbtidl/AGBT05B_047_01.summary
mpound, line 1: Why is this summary file added? Where is it used?

src/dysh/spectra/spectrum.py
astrofle: Good point. I was following the advice from `astropy` [here].(https://github.com/astropy/astropy/blob/75ce25ee3718cef767ca41dcbbf724a28bbe60a5/astropy/nddata/mixins/ndarithmetic.py#L593)

src/dysh/spectra/spectrum.py
```diff
@@ -567,6 +566,21 @@ def make_spectrum(cls, data, meta, use_wcs=True, observer_location=None):
         #
         return s
 
+    def __add__(self, other):
+        addition = self.add(other, **{"handle_meta": self._add_meta})
+        addition._target = self._target
```
mpound, line 571: What if other has a _target that does not = self._target?

testdata/AGBT05B_047_01/gbtidl/AGBT05B_047_01.summary
astrofle: I was using for testing the `summary` function.
I'll add the test.

mpound: Ah, I see that dysh-at uses it

src/dysh/spectra/spectrum.py
astrofle: My thinking is that we should preserve the attributes of the first operand when no changes are needed. This because one usually does `sig - ref` and wants the information of the signal.

src/dysh/spectra/spectrum.py
mpound: that astropy call goes thru [self._prepare_then_do_arithmetic() ](https://github.com/astropy/astropy/blob/75ce25ee3718cef767ca41dcbbf724a28bbe60a5/astropy/nddata/mixins/ndarithmetic.py#L657l) which checks for the None.  I think just add 

    if operand2 is None:
        meta["EXPOSURE"] = operand["EXPOSURE"]

mpound:
I don't understand why these tests are failing since they appear to be referencing changes I made in the 194-selection branch.

mpound (approved):
I don't know why the front page shows the old failing tests.   If you click on Checks tab, the tests are passing.    If you want me to do the merge let me know.

astrofle:
Thanks Marc. Go ahead with the merge. I don't know what is going on with the tests.

PR #218, opened 2024-03-26
197 q4 selection application

mpound:
This uses the Selection mechanism to calibrate PS scans.   It also modifies how we use the PSScan class based on discussion after the last release.   A PSScan now hold ONE scan pair, ONE IF, ONE feed, and multiple polarizations.    The previous version had multiple scan pairs. 

ScanBlock.timeaverage() has also changed so that it returns a Spectrum which is the average of all the PSScans it holds. Previously it returned an array of PScan.timeaverage()

The pytests have been updated accordingly. (Note there is one test where I had to lower the bar for Tsys difference with GBTIDL.   It is still quite small and just a matter of understanding how GBTIDL computes a (weighted?) averaged of an array of system temperatures.

A new notebook developer/test_getps_selection.ipynb  flexes the new capability.


mpound:
Note baseline() is still not fixed.  This pull is just about selection.

mpound:
Pytests should not be failing -- converted to draft till I figure this out.  You can still look at the code if you want to get a jump on this.

src/dysh/spectra/scan.py
```diff
@@ -194,7 +235,7 @@ def __init__(
         self._calstate = calstate  # ignored?
         self._scanrows = scanrows
         # print("BINTABLE = ", bintable)
-        # @todo deal with data that crosses bintables
+        # @todo deal with data that tcalcrosses bintables
```
astrofle, line 238: Looks like a typo here.

astrofle:
`pytest` is failing locally too. Maybe something was not committed?

Anyway, for `test_spectrum` the change is:
```python
self.ps0 = getps0.timeaverage()
self.ps1 = getps1.timeaverage()
```
instead of the current
```python
self.ps0 = getps0.timeaverage()[0]
self.ps1 = getps1.timeaverage()[0]
```

For `test_gbtfitsload` it requires changing the selection for picking the number of integrations in the `summary` method. I do not know how to fix this one and won't try.

mpound:
yeah, I think I may need to merge from main to my branch to get the arithmetic PR.  likely to be conflicts so I'll do this locally.

mpound:
Ok, I have fixed the tests that were failing due to API change.   We now have one truly failing test  test_getps_acs() where the assertion fails.   This is likely due to a difference in how GBTIDL and ScanBlock are doing time averages of PS scans.  In ScanBlock, we do an average of timeaverages using tsys weighting.  I suspect there is a weighting difference now that is causing a difference in the final time averaged spectrum.

astrofle:
> Ok, I have fixed the tests that were failing due to API change. We now have one truly failing test test_getps_acs() where the assertion fails. This is likely due to a difference in how GBTIDL and ScanBlock are doing time averages of PS scans. In ScanBlock, we do an average of timeaverages using tsys weighting but I also have to decide I suspect there is a weighting difference now that is causing a difference in the final time averaged spectrum.

Was the problem with test `test_getps_acs()` been fixed now? The CI shows it as passing.

etsmit:
I am getting an error in test_getps_selection.ipynb on cell 9:
image
image
I am also not seeing any empty curly brackets returned on cell 8. I can send the full traceback in slack/email.

etsmit:
After getting set up on a GBO redhat7 machine (python 3.9.16), it works without errors. It returns all the same warnings, which is fine, but I'm also not seeing some of the other print statements - such as the curly brackets printed in cells 8 and 11.

There may be something wrong with my local windows install (python 3.9.18), although I'm pretty sure I've got it set up correctly. Unless I can think of something else, I will probably nuke it and reinstall before the meeting at 4pm.

mpound:
The curly brackets are a leftover debug print of some method kwargs. I'll have to hunt that down.

mpound:
I can reproduce the no _spectral_axis error.  If you try to make a spectrum from another one, e.g.

s1 = ps.timeaverage()
s2 = Spectrum(s1)

You get the error at line 61:

  File ~/src/dysh/src/dysh/spectra/spectrum.py:61 in __init__
    self._spectral_axis._target = self._target

AttributeError: 'Spectrum' object has no attribute '_spectral_axis'
This appears to be related to the copy constructor for Spectrum1D

etsmit (approved):

PR #223, opened 2024-04-01
Total power using selection

mpound:
This converts gettp to use the selection mechanism. All tests pass. 
See also the bottom of developer/test_getps_selection.ipynb


mpound:
hmmm...why didn't this test fail in my sandbox previously?  I will fix it.

etsmit:
Do we understand the failing test in the 3.10 macos build? The weighted average unit test is returning a difference slightly larger than the threshold.

etsmit (approved):
- Do we understand why the macos 3.10 build is failing that one check?
- In the notebook, can we not select by sig and cal states (in the selection object) too?

mpound:
> In the notebook, can we not select by sig and cal states (in the selection object) too?

We can but because the CAL and SIG columns in SDFITS are chars, one would have to do, e.g.

sdf.select(cal='T')

We could add an issue for enhancement that select for any column that takes values 'T' or 'F' can use select(value=True/False).   I don't want to change the gettp() API to force cal and sig to be chars because that is less pythonic.

PR #225, opened 2024-04-02
Subbeamnod selection

mpound:
This PR implements selection mechanism for subbeam nod scans.
It also updates the example notebook examples/subbeamnod.ipynb to accomodate the changes to ScanBlock.
A new section is added in developer/test_getps_selection.ipynb to show the select() works for subbeamnod().

Also, some fixes were need in spectrum.py and spectralplot.py to restore attributes that went missing during arithmetic operations.

Also, I patched the macos failure we noted in the PR #223


etsmit (approved):
Note this invalidates the oldgetps() cells of test_getps_selection.ipynb.

mpound:
I will fix that oldgetps cell in a different PR.  Thanks.

Issue #229, opened 2024-04-18
Can't have xaxis unit as channels in SpectrumPlot


mpound:
**Feature description**
Plotting a spectrum by channel number

**Solution**
Need to peel off special case of _xaxis_unit='chan'_ in plot.

**Additional context**
The reason this fails is because there is no "channel" astropy Quantity, so we need to circumvent the conversion from Spectral units to the request xaxis_unit

astrofle:
Would defining a `"chan"` unit help? See e.g., https://docs.astropy.org/en/stable/units/combining_and_defining.html#defining-units

mpound:
That is a good idea, have to figure out how equivalencies() work with that.    We already have defined "mmHg" because it is not standard but in our FITS files.

mpound:
This is now fixed in release-0.3.0 and main.

PR #232, opened 2024-04-29
231 make an example notebook for doppler corrections

mpound:
This notebook shows how to change velocity frames and doppler conventions in spectra and in spectral plots.
This PR also fixes issue #229, which was caused by an failure to update a method signature of a non-public method when its usage changed.


mpound:
The notebooks is examples/velocity_frames.ipynb by the way

etsmit (approved):

Issue #256, opened 2024-05-23
ScanMixin does not understand drift scans


etsmit:
Summary: When trying to load in some test driftscan data, the timeaverage() function breaks as it doesn't understand the coordinate system when using the inherited astropy methods.

Recently, we tried to load in some test data that was taken during maintenance - while the telescope was locked in position. The point of this test data was to inject a test tone at a known frequency, take a short scan, and ensure it passes through the IF/LO system without drifting/shifting more than a few tens of Hz. This is done with the astrid command "Track(GetCurrentLocation("Encoder"), None, 30)" (see footnote). It essentially results in an SDFITS file with all blanks under the RADESYS column.

The file loads in OK as a GBTFITSLoad object, and I can retrieve e.g. "tpscan = sdf.gettp(scan=X,plnum=Y)" with no complaint. However, when running "tpscan.timeaverage()" I get the following traceback:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[128], line 1
----> 1 tpscan.timeaverage()

File ~/Git/dysh/src/dysh/spectra/scan.py:197, in ScanBlock.timeaverage(self, weights, mode)
    195 self._timeaveraged = []
    196 for scan in self.data:
--> 197     self._timeaveraged.append(scan.timeaverage(weights))
    198 if weights == "tsys":
    199     # There may be multiple integrations, so need to
    200     # average the Tsys weights
    201     w = np.array([np.nanmean(k._tsys_weight) for k in self.data])

File ~/Git/dysh/src/dysh/spectra/scan.py:512, in TPScan.timeaverage(self, weights)
    510 if self._npol > 1:
    511     raise Exception("Can't yet time average multiple polarizations")
--> 512 self._timeaveraged = deepcopy(self.total_power(0))
    513 if weights == "tsys":
    514     w = self._tsys_weight

File ~/Git/dysh/src/dysh/spectra/scan.py:492, in TPScan.total_power(self, i)
    490 restfreq = rfq.to("Hz").value
    491 meta["RESTFRQ"] = restfreq  # WCS wants no E
--> 492 return Spectrum.make_spectrum(self._data[i] * u.ct, meta, observer_location=self._observer_location)

File ~/Git/dysh/src/dysh/spectra/spectrum.py:534, in Spectrum.make_spectrum(cls, data, meta, use_wcs, observer_location)
    532     wcs = None
[...]

PR #264, opened 2024-05-30
Issue 256

astrofle:
fix: add support for AzEl observations

Adds check for empty `RADESYS` column and functionality to handle AltAz (AzEl) coordinate frame

Still needs tests, but it can handle Issue #256.


etsmit (changes requested):
Two things:
1. Add more to the warning (either at the GBTFITSLoad step or when changing velocity frames) that the Az/El values are invalid and will result in an incorrect frequency shift. This could be detected as a separate case from drift scans by detecting when e.g. elevation <4.9. (Or when az/el values are unchanging?)
2. Possibly add the scan numbers corresponding to the blank radesys rows to the warning message, since the antenna manager can be taken in/out over the course of a session. For example, "No RADESYS specified for scans X,Y,Z; assuming AltAz"

astrofle:
The last update also adds support for HaDEC observations.

src/dysh/fits/gbtfitsload.py
```diff
@@ -1633,3 +1635,28 @@ def write_scans(self, fileobj, scans, output_verify="exception", overwrite=False
...
+
+    def _update_radesys(self):
+        """
+        Updates the 'RADESYS' column of the index for cases when it is empty.
+        """
+
+        radesys = {"AzEl": "AltAz", "HADec": "hadec"}
+
+        warning_msg = (
+            lambda scans, a, coord, limit: f"""Scan(s) {scans} have {a} {coord} below {limit}. The GBT does not go that low. Any operations that rely on the sky coordinates are likely to be innacurate (e.g., switching velocity frames)."""
```
etsmit, line 1647: "inaccurate"

etsmit (approved):

astrofle:
@mpound is this fix acceptable?

mpound (approved):
Looks good to me.

PR #265, opened 2024-05-30
Baseline patch

teuben:
Here are a lot of improvements (with some caveats about units) for things related to comments on frequency switching for 0.3.0


astrofle:
@mpound @teuben Was there a decision made about the baseline selection units being displayed?

mpound:
@astrofle @teuben This is more than a displayed units issue.   It looks exclude_to_region() is not doing the conversion to SpectralRegion correctly in the case of input as channels.

astrofle:
Shouldn't this go into `release-0.3.0` ?

astrofle (approved):
Looks fine, but I think this should go into `release-0.3.0` not into `main`.

mpound:
It should go into both.

Issue #243, opened 2024-05-06
Numeric operations on Spectrum do not copy attributes


mpound:
**Describe the bug**
When e.g. multiplying a Spectrum by a value, Spectrum._shallow_copy_attributes is not invoked.   I think we have to define __rmul__ and __radd__

See also issue #241 
**How to Reproduce**
spectrum2 = 0.5*spectrum1
spectrum1.velocity_frame is None
{answer will be True}

**Environment**
- Dysh version 0.2.x
- Python version 3.9
- OS Ubuntu 24

PR #267, opened 2024-06-07
Fix: adds support for right hand side arithmetic operations on Spectrum

astrofle:
Adds `__rmul__`, `__radd__` and `__rsub__` operations for `Spectrum` objects and tests.
This should fix issue #243.


mpound (approved):
nice, thanks.

PR #273, opened 2024-06-13
create integration number when loading SDFITS data

mpound:
This PR adds _construct_integration_number to GBTFITSLoad.   It also renames _compute_proc to _construct_procedure_string and moves their invocation to _create_index_if_needed.  The move fixes a bug that @teuben noted when GBTFITSLoad(index=False).    A new SDFITS column INTNUM is added;  summary(verbose=True) includes INTNUM; and Selection no longer equates row with intnum (which as incorrect).   

A new test is added to compare INTUM with INT in a GBTIDL index file (added as a gzipped file which pandas can read directly).


mpound:
the original sdfits file was too big, and not actually in the repo.  I removed the gzipped index and added a smaller one (for A6)

PR #274, opened 2024-06-14
Update to ScanBlock documentation and implementation

etsmit:
Added a descriptive few sentences to the ScanBlock documentation. 

Removed the parent ScanMixin to better reflect usage of the ScanBlock implementation. For a scanblock containing many scans with varying numbers of polarizations, rows, nchan, etc. it is meaningless to return e.g. the ScanBlock.nchan attribute. Nothing else in ScanBlock currently relies on ScanMixin.


mpound:
ScanBlock does inherit the timeaverage(), polaverage(), finalspectrum() method interfaces, but you are right the rest doesn't make sense.   Possibly another Mixin class that just has those methods is appropriate, but that could be overkill.

mpound (approved):

etsmit:
Those functions are defined separately inside ScanBlock, instead of inheriting them from ScanMixin, right?

mpound:
Inherit in the sense that ScanMixin is a base class that derived classes can override methods.  Those methods in ScanMixin are just "pass" and each derived class defines them for itself.

Issue #269, opened 2024-06-11
GBTFITSLoad with index=False does not work


teuben:
**Describe the bug**
I wanted to see if index=False would speed up the loading. Instead I got an error:

```
  File ~/GBT/dysh/src/dysh/fits/gbtfitsload.py:142 in _compute_proc
    df = self._index["OBSMODE"].str.split(":", expand=True)

TypeError: 'NoneType' object is not subscriptable
```

**How to Reproduce**
```
f1 = util.get_project_testdata() / 'TGBT21A_501_11/TGBT21A_501_11.raw.vegas.fits'
sdf1 = GBTFITSLoad(f1,index=False)
```

My original file was 1.6GB in size, and it took 32sec to load, which seems long. That's an issue for another time.

**Environment**
- Dysh version 0.3
- Python version 3.11.5
- OS - kubuntu 22.04

we should make a dysh function to get these printed, including the branch name and git counter

Issue #250, opened 2024-05-10
Breaking data out of dysh


leemundy:
**Feature description**
What problem are you trying to solve that you currently can't?
Getting the frequency/velocity axis written into an astropy units object with my selection of units;
km/s in choice of reference frame, Hz, GHz, channel number etc. This can already be done for the
"y-axis" (flux).

**Solution**
If you have an idea of how to solve the problem, please describe it!

**Additional context**
The top level goal is to give the user their spectrum as arrays x and y that they can export from python
or use for whatever purpose in python.

leemundy:
Hi,

  If the person is familiar with python, this might mean creating local
variables with the spectra of interest, putting them
into their own dictionary(s) and writing out a pickle file. I agree that it
could also be an ascii file or FITS file. I would
argue that FITS is the least likely of the person just wants a spectrum to
be read into some modeling program,
or to put into another python notebook that makes a combined display of
various images and spectra -- for
publication.

Lee

mpound:
This is related to milestone "integrated read/write for SDFITS, Spectrum, ScanBlock" for the current quarter.

mpound:
This development now completed in the main branch.  You can save data from a Spectrum as ASCII of various flavors or FITS.  Also can save Scan/ScanBlock as SDFITS.  Also can subselect data from an SDFITS file and write it to a new SDFITS file.

Issue #227, opened 2024-04-17
Add virtual SDFITS columns to SDFITSLoad index


mpound:
**Feature description**
The primary HDU has metadata that should be made accessible as a virtual column in the index/selection object.  This is described in the design document. 

**Solution**
Quickest solution is just to make an additional column in the DataFrame with a constant value.    See DataFrame.assign
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html

mpound:
See also my comments on issue  #213

mpound:
A possible implementation is to write a custom __getitem__ for GBTFITSload, which would nicely allow the follwing

```
sdf = GBTFITSLoad(myfile)
sdf["FOOBAR"]    # returns column FOOBAR in the sdfits file or in the primary header
```

The pseudo-implementation is as follows

```
def __getitem__(self, key):
    try:
       return self._index[key]
    except KeyError:
       return self._primary_header[key]
```

mpound:
SDFITSLoad._add_primary_hdu() now does this

PR #276, opened 2024-06-24
270 readwrite interface for spectra scan scanblock

mpound:
This is a partial completion of the milestone.  This PR contains

1) Rewrite of SDFITS/GBTFITSLoad write method to allow selection and multifile output and test thereof
2) Spectrum read/write of FITS files, read/write of ESCV files, read of a variety of ASCII formats, read of GBTIDL ASCII format and tests thereof
3) Fixes issue #227 - Add primary HDU keywords to FITSload index
4) Fixing how integration numbers are created. (still one wrinkle to solve for SDFITS output)
5) Fixes issue #250 
6) Fixes issue #269 

What is still needed
- Output for Scan/ScanBlock  -- this entails essentially making and SDFITS format writer. 
- Documentation & notebooks


mpound:
The failures are because astropy 6 has changed the method name from update to update_header and we are pinned at 5.3 in these tests.  Hmmm

astrofle:
I think it would be better to have the multifits write mode append the VEGAS bank name rather than a number to the filenames.

mpound:
Yes, it would be better.   But the code should be general, so it would have to do something like grab the last char of the existing filename stem. But that does not cover the case of a project directory with FITS files that don't share the same naming convention.   This likely will need to be handled with an additional user argument to either follow the convention of the existing filesnames or use numbers....?

astrofle:
If you do
```Python
sdf = GBTFITSLoad(...)
sdf.write("foo.fits")
```
you get an exception about the selection resulting in zero rows.

mpound:
> If you do
> 
> ```python
> sdf = GBTFITSLoad(...)
> sdf.write("foo.fits")
> ```
> 
> you get an exception about the selection resulting in zero rows.

Fixed in latest push

mpound:
@astrofle The recent pushes have Scan/ScanBlock writing.

astrofle:
will check tomorrow

mpound:
Latest push has notebook examples.  I didn't put in a working example of multifits just noted it is possible because I did not want to force users to download multiple FITS files.

mpound:
also fixed the conflicts with main

astrofle:
Thanks Marc. I'll review it on Monday.

astrofle:
This causes an error:
```Python
from dysh import util
f = util.get_project_testdata() / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas/"
g = GBTFITSLoad(f)
g.write("merged.fits", multifile=False)
```
Error:
```Python
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
Cell In[1], line 5
      3 f = util.get_project_testdata() / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas/"
      4 g = GBTFITSLoad(f)
----> 5 g.write("merged.fits", multifile=False)

File /home/sandboxes/psalas/Dysh/dysh/src/dysh/fits/gbtfitsload.py:1745, in GBTFITSLoad.write(self, fileobj, multifile, verbose, output_verify, overwrite, checksum, **kwargs)
   1743         print(f"Total of {total_rows_written} rows written to files.")
   1744 else:
-> 1745     hdu = self._sdf[k]._hdu[fi[0]].copy()
   1746     outhdu = fits.HDUList(hdu)
   1747     for k in fi:

UnboundLocalError: local variable 'k' referenced before assignment
```

mpound:
Do a pull and try it now please

astrofle (approved):
Great work

PR #278, opened 2024-06-27
Baseline patch

teuben:


src/dysh/spectra/spectrum.py
mpound, line 1: Can the user provide both an include and an exclude or are they exclusive?

astrofle: Personally, I considered them exclusive. Give one or the other, not both.

notebooks/examples/frequencyswitch.ipynb
mpound, line 1: Cell 8 talks about exclusion region, but the example cell 9 uses inclusion.   Need to update #8.

src/dysh/spectra/core.py
mpound, line 1: It would be cleaner if the model selection was a dict pointing to a class, e.g. available_models = { "polynomial" : Polynomial1D, "chebyshev": Chebyshev1D, etc}  then selected_model = available_models[model](degree=order).  Gets rid of all the elifs and makes it easier to add additional models.   Also _valid_models could be replaced with list(available_models.keys())

src/dysh/spectra/spectrum.py
teuben: Meant to be exclusive, perhaps not well stated.

src/dysh/spectra/core.py
teuben: agreed, good suggestion. I have something similar for smoothing in two places, suggesting it should be placed in a common class.

src/dysh/spectra/spectrum.py
```diff
@@ -129,30 +155,60 @@ def baseline(self, degree, exclude=None, **kwargs):
...
         }
         kwargs_opts.update(kwargs)
 
+        if kwargs_opts["normalize"]:
+            print("Warning:  baseline fit done in channel space")
+            spectral_axis = np.copy(self._spectral_axis.value)
+            nchan = len(spectral_axis)
+            # sadly, there is no single setter for this
+            for i in range(nchan):
+                self._spectral_axis[i] = i * u.Hz  # would like to use "chan" units
```
mpound, line 193: Why not use u.chan ? 
Also you could do this to avoid the loop and not lose the other attributes (_target, _observer, etc) of spectral_axis
self._spectral_axis._value = np.arange(nchan)*u.chan
self._spectral_axis._unit = u.chan

src/dysh/spectra/spectrum.py
```diff
@@ -162,6 +218,9 @@ def _undo_baseline(self):
         if self._baseline_model is None:
             return
         if self._subtracted:
+            if self._normalized:
+                warnings.warn("Cannot undo previously normalized baseline subtraction")
```
mpound, line 222: Don't you already undo it in line 208?

src/dysh/spectra/spectrum.py
```diff
@@ -129,30 +155,60 @@ def baseline(self, degree, exclude=None, **kwargs):
...
         self._baseline_model = baseline(self, degree, exclude, **kwargs)
+
         if kwargs_opts["remove"]:
             s = self.subtract(self._baseline_model(self.spectral_axis))
             self._data = s._data
             self._subtracted = True
 
+        if kwargs_opts["normalize"]:
+            for i in range(nchan):
+                self._spectral_axis[i] = spectral_axis[i] * u.Hz
```
mpound, line 209: Line 209: What if the original unit wasn't Hz?    Need to save the unit or deep copy the entire spectral axis and restore it with self._spectral_axis = saved_spectral_axis

notebooks/examples/frequencyswitch.ipynb
teuben: Fixed docs on include/exclude.

The last cell I see is getting me in trouble, as it was background material what getfs() is doing.  It could be left out without loss of clarity, but we should have a notebook somewhere that goes more under the hood.

src/dysh/util/core.py
```diff
@@ -169,8 +196,34 @@ def get_size(obj, seen=None):
...
+    -------
+    string
+        matched string, if one is found.
+        Otherwise "None" is returned.
+
+    """
+    n = len(valid_strings)
+    m = []
+    for i in range(n):
+        if valid_strings[i].find(s) == 0:
```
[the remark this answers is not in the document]
teuben: I added a test like this to the pytest, but it's passing just fine.  I test on ==0, meaning it has to start at the string, so should be fine as it was.

src/dysh/spectra/spectrum.py
teuben: doesn't work for me:

```
UnitTypeError: SpectralAxis instances require units equivalent to '(Unit("Hz"), Unit("m"), Unit("J"), Unit("1 / m"), Unit("km / s"), Unit("pix"))', so cannot set it to 'chan'.
```

src/dysh/spectra/spectrum.py
teuben: It seems the _undo_baseline() function can be removed, as the removal is indeed done inside of baseline(). I doesn't hurt, since the self._subtracted was properly set. So perhaps somebody might want to not subtract, do some inspection, and later decide to call _undo_baseline() manually???

Must admit not having tested this.

src/dysh/spectra/spectrum.py
mpound: Oh, _value is not the correct attribute -- that does not change

src/dysh/spectra/spectrum.py
mpound: We want to keep _undo_baseline, in fact it should be promoted by removing the underscore.  It is the only undo we kept in the various requirements iterations.   It adds the model back in if subtracted and resets model to None.   You can remove lines 221/222 if self._normalized.

src/dysh/spectra/spectrum.py
teuben: I wind up doing a  deepcopy(), that works for now, minus the issue Hz/chan that needs to be resolved.

src/dysh/spectra/spectrum.py
```diff
@@ -205,15 +211,16 @@ def baseline(self, degree, exclude=None, include=None, **kwargs):
...
             del spectral_axis
+    # baseline
 
-    def _undo_baseline(self):
+    def undo_baseline(self):
         """
         Undo the most recently computed baseline. If the baseline
         has been subtracted, it will be added back. The `baseline_model`
         attribute is set to None. Exclude regions are untouched.
+        @todo    this function may seem redundant
```
mpound, line 224: Not sure why you think it is redundant, since the subtraction can be undone

mpound (approved):

teuben:
I still need to commit the example notebook, it was expanded a lot. but also ran into a bug where I could not take the average of two spectra. to be resolved.

PR #280, opened 2024-06-27
235 FFT shift

astrofle:
This implements the FFT shift method for folding frequency switched observations.
This method becomes the default, and the previous method can be accessed through the `shift_method='interpolate'` option.
It updates the tests to include both methods for shifting the spectra.


astrofle:
@teuben Could you review this? The PR is starting to have conflicts :/

src/dysh/fits/gbtfitsload.py
teuben, line 1: Shouldn't there a "The default is...." like the others do.
I prefer to all put them on a new line, but not sure yet if that survives formatting.

teuben (approved):

src/dysh/fits/gbtfitsload.py
astrofle: Thanks

PR #282, opened 2024-06-28
20m getps

astrofle:
Updates `GB20MFITSLoad`
* Updates the OBSERVER and OBJECT columns based on the header values.
* Updates the OBSMODE column removing non-ascii characters.
* Implements `getps` for onoff type of observations with the 20 Meter.
* Adds tests for `GB20MFITSLoad`


mpound (approved):
nice!

PR #283, opened 2024-07-01
Interactive plotter

vcatlett:


vcatlett:
I need to make these tests pass before I make this a full PR

PR #284, opened 2024-07-01
Smoothing1

teuben:
This adds smoothing to dysh. 

- the default decimation is decimate by the width parameter
- reference smoothing is implemented for gettp(smoothref=width), where width has to be odd if GBTIDL is to be reproduced (others have placeholders and need to be implemented)
-  hanning, boxcar and gaussian implemented using astropy.convolve
- Spectrum.smooth is normally called by the user, it returns a new Spectrum.
-  smoothref is done by calling the same service routine core.smooth that Spectrum.smooth also calls

```
Reduction of gaussian noise=1
width	boxcar  gaussian
3     	0.578  	0.470
7     	0.379  	0.308
11    	0.301  	0.244
21    	0.217  	0.177
31    	0.178  	0.145
```


astrofle:
Thanks Peter. Will try to get to it today, but it may be tomorrow.

astrofle:
In the smoothing notebook there's a comment about the correlation being lower when smoothing the reference spectrum during calibration and then smoothing, relative to smoothing the calibrated spectrum. I think this is because when smoothing the reference you are introducing a correlation between channels (since the channels in the reference spectrum are no longer uncorrelated). I can't think how to confirm this now.

teuben:
I think you are right, but there were a few other places I wasn't happy with the value of this correlation ratio not being what I expected. I may want to experiment with the decimation factor as well.

astrofle:
@teuben I left a PR with the changes I could find and added a test.
From my point of view what still needs to be done is update the example notebook so it works (add `numpy` import and smooth command in markdown cell at introduction).

astrofle:
The bug you mention in the updated frequencyswitch notebook when doing:
```Python
avg = ta2+ta3
avg = 0.5*avg
```
seems to have been solved in a recent merge (I could not reproduce it when running the above from `main`).

astrofle (approved):
Looks good. Thanks Peter!

teuben:
odd that you had issues with missing numpy. thanks for reviewing.

PR #290, opened 2024-07-04
allow BUNIT or TTUNITxx from the DATA column to carry their unit into…

teuben:
… the spectrum

Currently SDFITSLoad() always uses u.ct as the unit.   If BUNIT is in the bintable header, or TTUNITxx carries the unit (it should), then makespectrum is called with that unit..

It not, it will fall back to the classic u.ct

### DIFF
--- a/src/dysh/fits/sdfitsload.py
+++ b/src/dysh/fits/sdfitsload.py
@@ -436,18 +436,19 @@ class SDFITSLoad(object):
                 if v == "DATA":
                     ukey = "TUNIT" + k[5:]
                     break
-            for k, v, c in h.cards:
-                if k == ukey:
-                    if bunit != v:
-                        print("Found BUNIT=%s, now finding %s=%s, using the latter" % (bunit, ukey, v))
-                    bunit = v
-                    break
-        if bunit != None:
+            if ukey is not None:          # ukey should almost never be "None" in standard SDFITS
+                for k, v, c in h.cards:
+                    if k == ukey:
+                        if bunit != v:
+                            print("Found BUNIT=%s, now finding %s=%s, using the latter" % (bunit, ukey, v))
+                        bunit = v
+                        break
+        if bunit is not None:
             bunit = u.Unit(bunit)
         else:
             bunit = u.ct
 
-        s = Spectrum.make_spectrum(data * u.ct, meta, observer_location=observer_location)
+        s = Spectrum.make_spectrum(data * bunit, meta, observer_location=observer_location)
         return s
 
     def nrows(self, bintable):
--- a/src/dysh/fits/sdfitsload.py
+++ b/src/dysh/fits/sdfitsload.py
@@ -436,7 +436,7 @@ class SDFITSLoad(object):
                 if v == "DATA":
                     ukey = "TUNIT" + k[5:]
                     break
-            if ukey is not None:          # ukey should almost never be "None" in standard SDFITS
+            if ukey is not None:  # ukey should almost never be "None" in standard SDFITS
                 for k, v, c in h.cards:
                     if k == ukey:
                         if bunit != v:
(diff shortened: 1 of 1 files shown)


src/dysh/fits/sdfitsload.py
```diff
@@ -420,6 +420,33 @@ def getspec(self, i, bintable=0, observer_location=None):
...
+            for k, v, c in h.cards:
+                if k == "BUNIT":
+                    bunit = v
+            ukey = None
+            for k, v, c in h.cards:  # loop over the (key,val,comment) for all cards in the header
+                if v == "DATA":
+                    ukey = "TUNIT" + k[5:]
+                    break
+            for k, v, c in h.cards:
+                if k == ukey:
```
mpound, line 440: What happens here when ukey is None?

src/dysh/fits/sdfitsload.py
```diff
@@ -420,6 +420,33 @@ def getspec(self, i, bintable=0, observer_location=None):
...
+                    if bunit != v:
+                        print("Found BUNIT=%s, now finding %s=%s, using the latter" % (bunit, ukey, v))
+                    bunit = v
+                    break
+        if bunit != None:
+            bunit = u.Unit(bunit)
+        else:
+            bunit = u.ct
+
         s = Spectrum.make_spectrum(data * u.ct, meta, observer_location=observer_location)
```
mpound, line 450: should this now be data * bunit

teuben:
see also #295

src/dysh/fits/sdfitsload.py
```diff
@@ -420,6 +420,33 @@ def getspec(self, i, bintable=0, observer_location=None):
...
+                if k == ukey:
+                    if bunit != v:
+                        print("Found BUNIT=%s, now finding %s=%s, using the latter" % (bunit, ukey, v))
+                    bunit = v
+                    break
+        if bunit != None:
+            bunit = u.Unit(bunit)
+        else:
+            bunit = u.ct
+
```
mpound, line 449: I'd say we add meta["BUNIT"] if it doesn't exist.

src/dysh/fits/sdfitsload.py
teuben: it's fine (should be rare that TUNITxx is absent, but it would fall back to the "ct" unit, unless BUNIT in the header, then it will carry that across

src/dysh/fits/sdfitsload.py
teuben: yes, I swear that was in my editor, hope i didn't mess up branches.

mpound (changes requested):
see my comments

src/dysh/fits/sdfitsload.py
teuben: if will never exist, because at the moment meta only grabs from the columns.

src/dysh/fits/sdfitsload.py
mpound: Seems like you could skip the loop if ukey is None

src/dysh/fits/sdfitsload.py
teuben: sadly, I did mess up a branch. will need to run through tests again.

src/dysh/fits/sdfitsload.py
```diff
@@ -420,6 +420,33 @@ def getspec(self, i, bintable=0, observer_location=None):
...
+                if v == "DATA":
+                    ukey = "TUNIT" + k[5:]
+                    break
+            for k, v, c in h.cards:
+                if k == ukey:
+                    if bunit != v:
+                        print("Found BUNIT=%s, now finding %s=%s, using the latter" % (bunit, ukey, v))
+                    bunit = v
+                    break
+        if bunit != None:
```
mpound, line 445: should be 'if bunit is not None'.  ([PEP 8](https://peps.python.org/pep-0008/)

mpound (approved):

PR #300, opened 2024-07-10
Clean tests

astrofle:
Minor clean up of the tests to avoid keeping files created by I/O tests and to avoid failing tests when the GBO network is down.


mpound (approved):
Thanks for doing this.   In the future, I'd like to see us move to using a general teardown method common to unit testing.

PR #321, opened 2024-07-17
307 wcs wrong after a smoothed spectrum is indexed

teuben:
The stakeholders demo showed that the WCS for a decimated spectrum was not correct, although the freq. axis was correct. Although it was fixed by modifying the meta[] values, (only CRPIX1 was wrong) the interplay of wcs with meta and freq-axis needs to be
clarified in the meta discussion? Seems easy for developers to mess this up.   Ideally WCS updates need to occur in one place, we have several versions of this in spectrum.py, one for decimation, one for indexing a spectrum with single stride.


astrofle:
This does not handle the case where `decimate != width`.

For example:
```Python
import numpy as np

filename = get_project_testdata() / "TGBT21A_501_11/TGBT21A_501_11.raw.vegas.fits"

sdf = GBTFITSLoad(filename)
sb = sdf.getps(scan=152, ifnum=0, plnum=0)
ta = sb.timeaverage()
ts = ta.smooth('gaussian', 16.5, decimate=16)

# Check that CDELT1 is the channel increment in Hz.
cdelt1 = np.diff(ts.spectral_axis)
assert cdelt1[0].value == ts.meta["CDELT1"] # This fails.

# Check that CDELT1 is the original value times the decimation factor.
assert ts.meta["CDELT1"] == ta.meta["CDELT1"]*16 # This fails.

# Check that FREQRES has been updated and matches the original value times the smoothing width.
assert ts.meta["FREQRES"] == ta.meta["FREQRES"] * 16.5 # This fails.
```

astrofle:
The PR needs tests.

astrofle (approved):
After toying with the code in other places I see the need for a larger conversation around WCS and SDFITS definitions. I think we can keep this as is for now.

Issue #318, opened 2024-07-16
Q6 -  Metadata management


mpound:
**Feature description**
Metadata management as described in requirements. This is milestone for Q6

PR #323, opened 2024-07-22
replace index with selection

mpound:


mpound:
Previously GBTFITSLoad._index and _selection were separate but almost identical objects.    This PR gets rid of _index in favor of _selection.   I created a property _index for backwards compatibility.    There are still plenty of references to _index in GBTFITSLoad and in the rest of our code.    When I do issue #318, I will replace references to _index["KEY"] with self["KEY"]; so sense changing them to _selection["KEY"] now.

etsmit (approved):

PR #324, opened 2024-07-26
318 q6 metadata management

mpound:
This satisifes the metadata management requirement.  In this PR:
* Any SDFITS column can be accessed with sdf[colname], including DATA.
* colname is case insensitive, so e.g., sdf['ifnum'] works.
* Any SDFITS column can be set with sdf[colname] = whatever.   
* Slices of the data column can be set under certain conditions.  See below for note on DATA column
* Changes to metadata are written for both single and multifile objects
* Column names can be see with `column` property
* Tests of all functionality, including new file test_sdfitsload.py
* A new test file with multiple binary tables TGBT17A_506_11
* The user is warning if changing an existing column
* The user can add or delete a column.
* The user is not allowed to delete the DATA Column

**Note on setting the DATA column:** For data with multiple binary tables or multiple files, you can't set the data directly with sdf['data'] because np.vstack is called to concatenate the bintables or files and thus it is a new object.  For GBTFITSLoad with multiple files, you have to do the setting on the underlying SDFITSLoads.   e.g.,  g._sdf[0]["DATA"] = <whatever>.    For multi binary tables, one has to do sdf._bintable[n].data['DATA"] = <whatever>.    We can think about how to make this more user friendly.

Windows tests are failing because....Windows.


astrofle:
selection does not seem to be interacting with the metadata. That is, if I set a selection then the metadata getter still returns all of the metadata.

Example:
```Python

filename = get_project_testdata() / "AGBT05B_047_01/AGBT05B_047_01.raw.acs/AGBT05B_047_01.raw.acs.fits"

sdfits = GBTFITSLoad(filename)

# Select and then get metadata.
sdfits.select(scan=51, plnum=0)
sdfits["TCAL", "PLNUM", "SCAN"][0:1]
    
    	  TCAL	  PLNUM	  SCAN
0     1.424292	  1	  51
```

mpound:
There are two different selection/metadata interactions possible
1) Selection affects what sdfits[] does.   My contention is we don't want that.
2) Changing metadata via sdfits[] changes a previously selected selection.    I raised this in the design document but there was no discussion so no consensus.   At minimum changing metadata should warn that they have changed a column that was involved in a selection.  I have a code comment about this, but did not implement it.    Going further would be updating the selection based on the new metadata.   The user may not always want this, so we could/should provide an update_selection() method that would do this for them if they want it. 
How about I implement the warning, and we submit an enhancement issue for update_selection()?

mpound:
Also, the failing tests on windows are because windows is keeping a file handle open, so that write(overwrite=True) fails.   The workaround would be either to remove the file first in the test code (may not work for the same reason) or generate a new file name for each write in every test.  Ugh.

astrofle:
> There are two different selection/metadata interactions possible
> 
> 1. Selection affects what sdfits[] does.   My contention is we don't want that.
> 2. Changing metadata via sdfits[] changes a previously selected selection.    I raised this in the design document but there was no discussion so no consensus.   At minimum changing metadata should warn that they have changed a column that was involved in a selection.  I have a code comment about this, but did not implement it.    Going further would be updating the selection based on the new metadata.   The user may not always want this, so we could/should provide an update_selection() method that would do this for them if they want it.
>    How about I implement the warning, and we submit an enhancement issue for update_selection()?

Solution 2. sounds good to me.

mpound:
I have added the warning, a test for it, and created issue #327 for updating the selection.   We probably need a general solution for the windows issue as it is likely to come up in other contexts.

vcatlett:
The Windows tests appear to be failing with `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` upon attempting to run `fits.write(overwrite=True)`. Try opening the FITS file in "update mode" like `fits.open(filename.fits, mode='update')`, then running `hdul.flush()` instead of `hdul.writeto()`. Relevant Astropy documentation can be found here: https://docs.astropy.org/en/stable/io/fits/#save-file-changes

mpound:
windows checks now pass.

astrofle (approved):
Could not break it :)
Excellent work!

PR #331, opened 2024-07-31
Change docs theme and add notebook rendering

vcatlett:
Here's what I did:
- Changed the Sphinx theme from `sphinx-rtd-theme` to `sphinx-book-theme`
- Added Jupyter Notebook rendering via `myst_nb` (which replaces `myst_parser`)
- Added theme option in `conf.py` to cache notebook executions so they only re-run if a cell gets changed (very useful during autobuilding)
- Moved notebook inputs/outputs to subfolders of `notebooks/examples/` as not to pollute that folder upon a docs build
- Changed `docs/source/examples/` to be a symlink to `notebooks/examples/` so we don't have to duplicate work. Note that Windows users have to manually create the symlink with admin permissions. This does not affect unit tests or the ReadTheDocs build
- Added an `index.rst` file to `notebooks/examples/` so Sphinx can generate the appropriate links and table of contents
- Added separate logos for light and dark mode
- Updated `pyproject.toml` and `requirements.txt` to reflect the new dependencies
- Updated `.gitignore` to ignore various docs directories

The docs can be previewed here: https://dysh.readthedocs.io/en/cat-docs-devel/


astrofle:
Thanks a lot Cat.

I noticed this adds the outputs from the notebooks to the repo (e.g., testwrite.ascii). Do we need to do this?
I think it would be better to avoid adding these files unless we have to.

docs/source/conf.py
vcatlett, line 1: Note that `html_theme_options = {"logo_only": True, "vcs_pageview_mode": "display_github"}` is replaced with only logo theme options because `logo_only` and `vcs_pageview_mode` are not valid for this new theme.

docs/source/examples
vcatlett, line 1: This is the symlink. It seems to behave properly on GBO machines (RHEL7/8) and in the ReadTheDocs build. Windows machines interpret this as a text file and need to manually set the link.

vcatlett:
> Thanks a lot Cat.
> 
> I noticed this adds the outputs from the notebooks to the repo (e.g., testwrite.ascii). Do we need to do this? I think it would be better to avoid adding these files unless we have to.

My bad. I had the wrong path in the `.gitignore` and wasn't paying close attention. `notebooks/examples/output/` should now be empty save a `.gitkeep` file (to keep the directory visible for RTD). `notebooks/examples/data/` should now only have 2 files, which are used by `dataIO.ipynb` but not downloaded by the notebook.

mpound:
I ran `hatch run docs-build` from a clean branch.   The frequency switch notebook example output has many cells with the download status bar for the example data: 
fsblech

How do I force a re-run to fix this?  I tried touching frequencyswitch.ipynb but that did not work.

teuben:
so either you run it twice, so it's there, or you point $DYSH_DATA to the tree where example_data is located. For this the dysh_data branch needs to be merged (but it's ready for a merge, pending discussion?)

mpound:
Running the docs build twice doesn't work because its sees the notebook is cached.   I tried removing what I thought was the cached version but that made no change.

vcatlett:
@mpound Curious that this doesn't happen in the ReadTheDocs build. There are two options I'll try. One is that in `conf.py`, I can add an option to suppress terminal output, although I worry it would suppress ALL output. The other is that `wget` should have an option to suppress the status bar. Stay tuned for a new commit.

vcatlett:
Latest commit includes:
- Dependency changes: Added `httpx` and `rich`, removed `wget`
- New method: `dysh.util.download.from_url()` downloads data from a given URL.  At @tchamberlin 's suggestion, it uses `httpx` with a `rich` progress bar instead of `wget`. 
- New test: Replaced `util/tests/test_wget.py` with `util/tests/test_download.py`
- Notebook changes: Replaced download code in each of the existing notebooks with the method mentioned above
- Notebook changes: Removed unused imports 
- Bug: The notebooks render an extra empty cell after the progress bar after these changes. It's still a lot better than the massive scrolling output in the previous commit

Note that I have not fixed the Sphinx warnings of the format `dysh/src/dysh/util/selection.py:docstring of dysh.util.selection.Selection:94: CRITICAL: Unexpected section title.` We should go back and fix all formatting warnings at some point, but it's not crucial at this stage.

vcatlett:
@mpound and @astrofle May I move the remaining input files to `/home/dysh/` so they get downloaded by the notebooks instead of needing to stay in the git repo?

This will move these files...
- `dysh/notebooks/examples/data/onoff-L_getps_152_RADI-HEL.ascii.gz`
- `dysh/notebooks/examples/data/onoff-L_gettp_156_intnum_0_HEL.ascii`

to here on the GBO file system...
- `/home/dysh/example_data/onoff-L/gbtidl-data/onoff-L_getps_152_RADI-HEL.ascii.gz`
- `/home/dysh/example_data/onoff-L/gbtidl-data/onoff-L_gettp_156_intnum_0_HEL.ascii`

which could then be downloaded from here:
- https://www.gb.nrao.edu/dysh/example_data/onoff-L/gbtidl-data/onoff-L_getps_152_RADI-HEL.ascii.gz
- https://www.gb.nrao.edu/dysh/example_data/onoff-L/gbtidl-data/onoff-L_gettp_156_intnum_0_HEL.ascii

Finally, I would edit `dysh/notebooks/examples/dataIO.ipynb` to download the files before reading them

vcatlett:
Moving the download message in `dysh.util.download.from_url.py()` disn't fix the extra cells, but it does make the printing order make more sense:
image

mpound:
@vcatlett Yes, you can move those files.   The copies of them that are in dysh/testdata should remain because they are used in pytests.

mpound:
Thank you for fixing the notebook download cells.
The code in download.py has similar functionality as in the dysh_data branch by @teuben .  See https://github.com/GreenBankObservatory/dysh/blob/dysh_data/src/dysh/util/files.py
These will need to be merged, but that can be done after this PR.

mpound:
The formatting of the notebook cells is inconsistent as you have noted.   For instance the use of header levels in markdown cells, which causes extraneous entries in the doc table of contents.    A new github issue should be made so we don't forget that.

vcatlett:
As far as I can tell, this PR is ready to be merged. I reported GitHub issues for the lingering bugs:
- https://github.com/GreenBankObservatory/dysh/issues/334
- https://github.com/GreenBankObservatory/dysh/issues/335
- https://github.com/GreenBankObservatory/dysh/issues/336

The docs are available here: https://dysh.readthedocs.io/en/cat-docs-devel/index.html

vcatlett:
My latest commits include:
- Merging @mpound's PR changes into this branch. 
- A first pass at fixing the header issues mentioned in https://github.com/GreenBankObservatory/dysh/issues/334
 
Note that I am leaving that bug open until:
- Notebook authors go through their notebooks to make sure the header hierarchy matches what they want
- We add a unit test to make sure future notebooks have exactly one top-level header

vcatlett:
> so either you run it twice, so it's there, or you point $DYSH_DATA to the tree where example_data is located. For this the dysh_data branch needs to be merged (but it's ready for a merge, pending discussion?)

@teuben My first-pass download function follows recommendations by @tchamberlin, namely replacing the `wget` dependency with `httpx`. Happy to review your PR when it's ready. You are welcome to overwrite my `dysh.util.download` module with your own provided it makes that dependency change and gets called by the notebooks without re-introducing the over-printed progress bar.

vcatlett:
Waiting to merge unit tests from this PR: https://github.com/GreenBankObservatory/dysh/pull/340

vcatlett:
> * https://www.gb.nrao.edu/dysh/example_data/onoff-L/gbtidl-data/onoff-L_getps_152_RADI-HEL.ascii.gz

These changes have been made. GBO computing had just accidentally set `/home/dysh/` to "read-only" but fixed that for us.

astrofle:
Why not use [`pytest-notebook`](https://pytest-notebook.readthedocs.io/en/latest/) for testing the notebooks?

vcatlett:
> Why not use [`pytest-notebook`](https://pytest-notebook.readthedocs.io/en/latest/) for testing the notebooks?

Because I didn't know about it. I can look into it, although since what I wrote already works, I'd rather just leave it unless there's a good reason to switch.

vcatlett:
Latest commits include:
- Cleaned up some old developer documentation
- Added instructions on how to use `pre-commit`
- Added instructions on how to update project requirements
- Fixed the "Starting download" message so it prints before the progress bar

astrofle:
> > Why not use [`pytest-notebook`](https://pytest-notebook.readthedocs.io/en/latest/) for testing the notebooks?
> 
> Because I didn't know about it. I can look into it, although since what I wrote already works, I'd rather just leave it unless there's a good reason to switch. (A good reason could be that this would do individual tests for each notebook rather than one big test that runs through all of them.)

Makes sense. We can try to switch after this PR is done.

astrofle (approved):
Thanks a lot for all the work you put into this.

vcatlett:
Latest changes include:
- Merged Pedro's changes from https://github.com/GreenBankObservatory/dysh/pull/350
- Re-compiled `requirements.txt` to reflect dependency changes (removing `glob2`)

mpound (approved):
@vcatlett This was a huge effort and I really appreciate all the great work you did!

PR #333, opened 2024-08-02
Cat docs devel mwp review

mpound:
Here are some small changes. Otherwise the PR looks great!


vcatlett:
Looks good to me. I'll merge it.

PR #345, opened 2024-08-08
Issue 256 index update

astrofle:
This is an ugly fix for issue #256.
It adds updating the "RADESYS" column of the underlying `SDFITSLoad.index` objects in `GBTFITSLoad._sdf`.

### DIFF
--- a/src/dysh/fits/gbtfitsload.py
+++ b/src/dysh/fits/gbtfitsload.py
@@ -1852,18 +1852,18 @@ class GBTFITSLoad(SDFITSLoad):
         # Update SDFITSLoad.index.
         sdf_idx = set(self["FITSINDEX"][azel_mask])
         for i in sdf_idx:
-            sdfi = self._sdf[i]
+            sdfi = self._sdf[i].index()
             azel_mask = (sdfi["CTYPE2"] == "AZ") & (sdfi["CTYPE3"] == "EL")
-            sdfi["RADESYS"][azel_mask] = radesys["AzEl"]
+            sdfi.loc[azel_mask, "RADESYS"] = radesys["AzEl"]
 
         # Hour angle and declination case.
         hadec_mask = self["CTYPE2"] == "HA"
         self._index.loc[hadec_mask, "RADESYS"] = radesys["HADec"]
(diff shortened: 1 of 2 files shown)


src/dysh/fits/gbtfitsload.py
```diff
@@ -1847,11 +1847,23 @@ def _update_radesys(self):
...
+            azel_mask = (sdfi["CTYPE2"] == "AZ") & (sdfi["CTYPE3"] == "EL")
+            sdfi.loc[azel_mask, "RADESYS"] = radesys["AzEl"]
 
         # Hour angle and declination case.
         hadec_mask = self["CTYPE2"] == "HA"
         self._index.loc[hadec_mask, "RADESYS"] = radesys["HADec"]
+        sdf_idx = set(self["FITSINDEX"][hadec_mask])
+        for i in sdf_idx:
+            sdfi = self._sdf[i].index()
+            azel_mask = sdfi["CTYPE2"] == "HA"
```
mpound, line 1865: for consistency with the previous loop, this variable should be hadec_mask.  doesn't change the functionality at all though

src/dysh/fits/gbtfitsload.py
```diff
@@ -1847,11 +1847,23 @@ def _update_radesys(self):
 
         # Azimuth and elevation case.
         azel_mask = (self["CTYPE2"] == "AZ") & (self["CTYPE3"] == "EL")
+        # Update self._index.
         self._index.loc[azel_mask, "RADESYS"] = radesys["AzEl"]
+        # Update SDFITSLoad.index.
+        sdf_idx = set(self["FITSINDEX"][azel_mask])
+        for i in sdf_idx:
+            sdfi = self._sdf[i].index()
```
mpound, line 1855: `sdfi = self._sdf[i]`
should ould also work since SDFITSLoad has __getitem__

mpound (approved):

src/dysh/fits/gbtfitsload.py
astrofle: In this case the update would require a chained indexing, which will be deprecated by `pandas`.

Issue #316, opened 2024-07-15
SubBeamNodScan has no `_nrows` attribute


astrofle:
**Describe the bug**
It is not possible to retrieve the `nrows` attribute of a SubBeamNodScan.

**How to Reproduce**
```Python
filename = "/home/dysh/example_data/subbeamnod/data/AGBT13A_124_06/AGBT13A_124_06.raw.acs/"
sdf_sbn = GBTFITSLoad(filename)
sbn_scan = sdf_sbn.subbeamnod(ifnum=0, plnum=0)
sbn_scan[0].nrows
```

**Environment**
- Dysh version 0.3.0
- Python version 3.11.9
- OS RedHat 7

Issue #58, opened 2023-07-21
Test subbeamnod using synthetic data


astrofle:
Since `GBTIDL` did not have a procedure to calibrate `SubBeamNod` scans we will have to use synthetic data for testing. 
We need to make a realistic SDFITS file for the test.

Issue #133, opened 2023-11-22
`gettp` use of cal and calibrate is inconsistent


astrofle:
When calling `gettp` and using `cal=True`, the returned data is still a combination of the `cal='T'` and `cal='F'` rows.

An example:

``` Python
import numpy as np

from dysh.fits.gbtfitsload import GBTFITSLoad
import dysh.util as util

testdata = util.get_project_testdata()
data_path = testdata  / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas/AGBT18B_354_03.raw.vegas.A.fits"

sdf = GBTFITSLoad(data_path, verbose=False)
tps = sdf.gettp(6, ifnum=2, cal=True, plnum=0)

# This should return True -- it does not.
np.all(tps[0]._refcalon == tps[0].total_power(0).flux.value)

# This should return False -- it does not.
np.all(0.5*(tps[0]._refcalon + tps[0]._refcaloff) == tps[0].total_power(0).flux.value)
```

mpound:
I find the GBTIDL documentation about cal_state and sig_state confusing:
https://www.gb.nrao.edu/GBT/DA/gbtidl/release/user/guide/gettp.html

It seems like regardless of cal_state/sig_state,  dototalpower is called to average both switching phases cal_state=0/1 ('F'/'T' in SDFITS header) and calculate Tsys.  But then some different average is done (keeping the original Tsys?) depending on cal_state & sig_state being 0,1, or unset.  

Specifically, I am having trouble parsing this sentence in the GBTIDL **Summary** of gettp:

_**Note:** this step is done even when the data from only one state is requested using the cal_state or sig_state keywords._

Here is my best guess at the expected behavior,  with Python representation 0=False, 1=True, None=unset:

 | CAL     |   SIG    |   RESULT |     TSYS
 | -----      |  -------   |  --------    |    ----  | 
 |  None  | None |    data = 0.5* (REFCALON + REFCALOFF) , regardless of sig state  | use all CAL states, all SIG states |
 | None  | True |      data = 0.5* (REFCALON + REFCALOFF), where sig = 'T'  (sig_state=1) |  use all CAL states, SIG='T' |
| None  | False |     data = 0.5* (REFCALON + REFCALOFF), where sig = 'F'  (sig_state=0) |  use all CAL states, SIG='F' |
| True   |   None |    data = REFCALON, regardless of sig_state | use all CAL states, all SIG states |
[...]

PR #352, opened 2024-08-20
SubBeamNod update

astrofle:
In this PR
* Fix for issue #316.
* Fix for `GBTFITSLoad.subbeamnod` when using `method='cycle'` after the updates to `TPScan`.
* Refactor of the code for `SubBeamNodScan` to use a single processing regardless of `method`.
* Adds tests using synthetic data -- issue #58.


src/dysh/spectra/scan.py
```diff
@@ -1543,8 +1543,6 @@ class SubBeamNodScan(ScanMixin):
         Signal total power scans
     reftp:  list ~spectra.scan.TPScan
         Reference total power scans
-    fulltp:  ~spectra.scan.TPScan
-        A full (sig+ref) total power scans, used only for method='scan'
     method: str
```
mpound, line 1546: method argument is no longer needed/used, right?

src/dysh/spectra/tests/test_scan.py
```diff
@@ -183,6 +180,85 @@ def test_compare_with_GBTIDL(self, data_dir):
...
+        # Convert to counts.
+        p_sig_mod = gain * (sig_mod + tsys)
+        p_ref_mod = gain * (ref_mod + tsys)
+
+        # Replace data.
+        sdf["TCAL"] = tcal
+        new_data = np.empty_like(sdf["DATA"])
+        new_data[sig_mask] = p_sig_mod
+        new_data[ref_mask] = p_ref_mod
+        sdf["DATA"] = new_data
```
mpound, line 230: I expect this generates a warning.   You could filter the warning beforehand.  Or not.

src/dysh/spectra/tests/test_scan.py
```diff
@@ -183,6 +180,85 @@ def test_compare_with_GBTIDL(self, data_dir):
...
+        assert pytest.approx(sbn_scan.data.mean(), rms_scan.value) == tcont
+
+        # Compare exposure times.
+        assert sbn_cycle.meta["EXPOSURE"] == sbn_scan.meta["EXPOSURE"]
+
+        # Compare system temperature.
+        assert pytest.approx(sbn_scan.meta["TSYS"], rms_scan.value) == sbn_cycle.meta["TSYS"]
+        assert pytest.approx(sbn_cycle.meta["TSYS"] - tcal / 2.0, rms_cycle.value) == tsys
+        assert pytest.approx(sbn_scan.meta["TSYS"] - tcal / 2.0, rms_scan.value) == tsys
+        # The need to re-subtract the calibrator temperature suggests that there is a problem
```
mpound, line 252: TPScan.calc_tsys() was re-written as part of #133.   Maybe a problem there?

mpound (approved):
Some minor suggestions.  Nice work!

src/dysh/spectra/tests/test_scan.py
astrofle: The base implementation of `spectrum.core.mean_tsys` did not change, and the remaining changes do not seem to affect the calculation itself, or other tests involving `gettp`. I think this may be a problem with the equation itself and adding back the `tcal/2.0` (in `spectrum.core.mean_tsys`), because the line amplitude also comes back higher than what you put in. I'd like to double check.

mpound: Tsys computation reference: https://www.gb.nrao.edu/GBT/DA/gbtidl/gbtcal2.1-30Oct09.pdf

astrofle: Yes, not `dysh` specific.

src/dysh/spectra/tests/test_scan.py
astrofle: Good idea.

src/dysh/spectra/scan.py
astrofle: Correct. Removed.

src/dysh/spectra/tests/test_scan.py
astrofle: My mistake. Nothing wrong with how things are.

PR #358, opened 2024-08-22
348 q6 logging of user commands for data reproducibility

mpound:
History logging of dysh commands.

- Adds HistoricalBase abstract class for managing history and comments (a better name might be CommentaryClass since FITS refers to HISTORY and COMMENTS as commentary cards. But I like Historical)
- Adds dysh history logger dhlogger 
- Adds decorators @log_call_to_history and @log_function_call.   I have not fully populated all the places these should be called by default.
- Refactors Scan and Scanblock classes to inherit from new ScanBase, SpectralAverageMixin, HistoricalBase as appropriate

I have not written any test code yet specific to HistoricalBase usage, though various read/write function tests imply that HISTORY and COMMENTS are preserved.


mpound:
This is a draft for you to peruse while I finish writing tests and any outstanding small changes.

mpound:
Not sure what's going on with the conflicts.   I merged main into my branch before committing.

mpound:
@astrofle @tchamberlin 
Some things to check on this:
1) Do calibration routines add history as you expect?
2) Is that history passed on correctly to data products (Spectrum, Scan, ScanBlock)?
3) Are you happy with the format of the history strings in sdf.history etc?
4) Can you add history and comments to GBTFITSLoad, Spectrum, Scan, ScanBlock?
5) Are history and comments output correctly on write? 
6) Do the decorators @log_call_to_history and @log_function_call work as you expect?
7) Try to break things as usual

For 4 and 5, I have pytests that are not yet checked in because of one error I'm tracking down.

astrofle:
Looking good. My comments thus far:
* The format for history looks good. Maybe comments should also log the time the comment was created?
* The history should be ordered by time.
* `ScanBlock` is not getting the history of how it was created. E.g., `dysh.fits.gbtfitsload.GBTFITSLoad.getps(scan=152,ifnum=0,plnum=0,)` is not in the `ScanBlock` history.
* After writing and loading a `ScanBlock` the history and comments are not preserved, i.e. the loaded `ScanBlock` has a new history without the history of the original `ScanBlock`.
* `Spectrum` has an empty history and comments.

mpound:
@astrofle All your comments have now been addressed except logging the time of the comment.   I chose not to do that because:
1) Looking at various FITS files, comments usually don't have times in them.  Comments imported on read would not have timestamps so why add them to dysh comments?
2) The timestamp in history is actually added by the logger for decorated calls.  Comments do not go through the logger nor do history cards that are added through add_history (i.e., user-added)

We could add an add_timestamp boolean option to add_history(), add_comment() that defaults to false.  I suggest we see if users request such timestamping.

mpound:
One thing I will note is that the decorator @log_function_call changed to take a log level, e.g. @log_function_call("DEBUG").  The default is "INFO".  However, a side-effect is that if you want the default INFO level the decorator must have empty parens, e.g '@log_function_call()'.     I tried a couple  things to "fix" this but could not get it to work.  So left as is.

astrofle:
Would it help if the `GBTFITSLoad` and `SDFITSLoad` objects did not keep track of steps that did not modify them?
For example, using `gettp` should not change the object, so it is not relevant to reproduce the data reduction.

astrofle:
Something neat you can do:
```Python
sdf = GBTFITSLoad(filename)
eval(sdf.history[-1].split(":")[-1].replace("dysh.fits.gbtfitsload.GBTFITSLoad", "sdf"))
```
it could be used to reproduce  data reductions without minimum user input.

astrofle:
I just noticed that the timestamps for the `ScanBlock` history are all the same. Not ideal.

mpound:
- The format for history looks good. Maybe comments should also log the time the comment was created?
   
add_time boolean argument added to add_history() and add_comment().  Defaults
 to false.

- The history should be ordered by time.
  
Done

- ScanBlock is not getting the history of how it was created. E.g., dysh.fits.gbtfitsload.GBTFITSLoad.getps(scan=152,ifnum=0,plnum=0,) is not in the ScanBlock history.

 Fixed the general problem of result classes not getting the function call that created them (e.g. getps). New decorator @log_call_to_result handles these cases.

- After writing and loading a ScanBlock the history and comments are not preserved, i.e. the loaded ScanBlock has a new history without the history of the original ScanBlock.

- Spectrum has an empty history and comments.

Both Fixed

- When time averaging a ScanBlock the history gets printed.

- The history of a ScanBlock does not say how it was created, and sometimes it contains history not associated with it:

- the above behavior also results in multiple history cards from previous ScanBlocks going into the last ScanBlock created.

Both of these fixed with @log_call_to_result

- I just noticed that the timestamps for the ScanBlock history are all the same. Not ideal.
  
I can't reproduce this.  There are two that are the same because the
  were written less than 1 second apart (the call to getps and the call
  to `Scanblock.__init__`).  Unless we want to include milliseconds in the
  time stamp, which I think is overkill, this is going to happen sometimes.

Also, added method to ensure that FITS cards being written out don't have non-ASCII characters which was a problem that cropped up after the above changes. 

@astrofle Please re-review

astrofle (approved):
Great work!

Issue #365, opened 2024-08-27
`Spectrum` arithmetic does not preserve the spectral axis attributes


astrofle:
**Describe the bug**
After doing any arithmetic operation on a `Spectrum` the result has a `spectral_axis` without `observer`, `target`, etc...

PR #368, opened 2024-08-29
Issue 365

astrofle:
In this PR:
* A fix for #365 
* A test that would have caught the issue in the first place

The solution was to `deepcopy` the attributes that made sense to be copied, i.e., those attributes that should not change after an arithmetic operation.
A caveat is that this does not check for on/off source or anything like that, it just copies the attributes from the first operand (when there are two or more `Spectrum` in the operation).
I juts put a "global" variable in spectrum.py to capture which attributes should be ignored, but please let me know if there's a better way of handling this. (I wanted to have access to the attributes that are not copied for the test.)


mpound:
Smart to use a list of 'don't copy' rather than a list of copy.  Helps with future maintenance.   
The other option for the variable IGNORE_ON_COPY would be to make it a class static variable of Spectrum.   I think it is fine how you have done it.

mpound (approved):

PR #371, opened 2024-09-03
Issue 236

astrofle:
In this PR:
* Changes to `make_spectrum` to avoid assuming that all the keywords are present.
* Attaches zero velocities to the `SpectralCoordinate` for the observer before validating to avoid a warning.

### DIFF
--- a/src/dysh/spectra/spectrum.py
+++ b/src/dysh/spectra/spectrum.py
@@ -9,7 +9,7 @@ import astropy.units as u
 import numpy as np
 import pandas as pd
 from astropy.coordinates import SkyCoord, SpectralCoord, StokesCoord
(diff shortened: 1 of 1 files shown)


src/dysh/spectra/spectrum.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 import pandas as pd
 from astropy.coordinates import SkyCoord, SpectralCoord, StokesCoord
-from astropy.coordinates.spectral_coordinate import NoVelocityWarning
+from astropy.coordinates.spectral_coordinate import attach_zero_velocities
```
mpound, line 12: Wow, didn't know there was such a function!

mpound (approved):
good job

Issue #338, opened 2024-08-04
ReadTheDocs raises new warnings


astrofle:
**Describe the bug**
There are some warnings that only appear in the ReadTheDocs build, for example in the metadata notebook on the baseline correction with the unmodified metadata ([this section](https://dysh.readthedocs.io/en/cat-docs-devel/examples/metadata_management.html#metadata-inspection)) there is the following:

```Python
/home/docs/checkouts/readthedocs.org/user_builds/dysh/envs/cat-docs-devel/lib/python3.9/site-packages/astropy/units/core.py:2045: ResourceWarning: unclosed <ssl.SSLSocket fd=53, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('172.17.0.2', 41644), raddr=('192.33.116.7', 443)>
  if hasattr(s, "_get_physical_type_id"):
```
which is not on the "plain" notebook.

**How to Reproduce**
Look at : https://dysh.readthedocs.io/en/cat-docs-devel/examples/metadata_management.html#metadata-inspection
from PR #331.

**Environment**
ReadTheDocs

astrofle:
I saw the same error locally. It seems this comes from the `from_url` function? (just guessing because it mentions addresses).

astrofle:
I think the problem is that the `httpx.Client` is not closed after the download. 
I tried reproducing the error by downloading multiple times 
```Python
for i in range(10):
    filename = from_url(url, savepath)
    filename.unlink()
filename = from_url(url, savepath)
```
and then doing other things in the code. The warning showed up four times in a row.

After adding `httpx.Client.close()`, the warning did not show up anymore.

PR #387, opened 2024-09-21
Fix: close httpx.client after download

astrofle:
Fixes issue #338 by closing the `httpx.client` after the download is finished.


vcatlett (approved):
Good catch, thanks.

Issue #361, opened 2024-08-23
gettp() does not separate the IF's in a nodding example


teuben:
**Describe the bug**
gettp() returns a single scanblock, even if no ifnum= is specified when there are multple IF's.

**How to Reproduce**
```
# from acceptance_testing/data/
sdf=GBTFITSLoad('TSCAL_19Nov2015/TSCAL_19Nov2015.raw.acs/TSCAL_19Nov2015.raw.acs.fits')

sdf.gettp(scan=6,fdnum=0,plnum=0)
sdf.gettp(scan=6,fdnum=0,plnum=0,ifnum=0)
sdf.gettp(scan=6,fdnum=0,plnum=0,ifnum=1)
```
each returns a single scanblock, but the first one should return 2.

It seems, only the first is returned in this case:

```
print(sdf.gettp(scan=6,fdnum=0,plnum=0)[0].timeaverage().stats())
print(sdf.gettp(scan=6,fdnum=0,plnum=0,ifnum=0)[0].timeaverage().stats())
print(sdf.gettp(scan=6,fdnum=0,plnum=0,ifnum=1)[0].timeaverage().stats())

(<Quantity 3.10215706 ct>, 1.3773813265212864, <Quantity 0.03795753 ct>, <Quantity 5.53300124 ct>)
(<Quantity 3.10215706 ct>, 1.3773813265212864, <Quantity 0.03795753 ct>, <Quantity 5.53300124 ct>)
(<Quantity 2.67228504 ct>, 1.078213412987175, <Quantity 0.0481095 ct>, <Quantity 12.03663866 ct>)
```

**Environment**
- Dysh version  0.3
- Python version 3.11.7
- OS linux/ubuntu 22.04

teuben:
And to come back to an earlier point, the following example shows POL is special.  
```
sdf.gettp(scan=6,fdnum=1,plnum=1,ifnum=1)[0].fdnum    #ok
sdf.gettp(scan=6,fdnum=1,plnum=1,ifnum=1)[0].plnum    #  fails
sdf.gettp(scan=6,fdnum=1,plnum=1,ifnum=1)[0].ifnum    #ok
```

mpound:
ScanBlocks can have different IFs, Scans cannot.  This is not technically a bug.  But we have to decide if we want to change the behavior of ScanBlock

mpound:
See also #332

astrofle:
[...]

PR #388, opened 2024-09-24
Fix: gettp does not separate IFs

astrofle:
Fixes issue #361


mpound (approved):
Good catch!

astrofle:
This touches `GBTFITSLoad` so I'll wait for Peter's nodding2 PR to merge (or we can try something else).

Issue #390, opened 2024-09-25
Missing INTNUMs


astrofle:
**Describe the bug**
When checking the integration numbers (column INTNUM) for a `GBTFITSLoad` object, for one of the spectral windows INTNUM starts at 12 even though it should start at 0.

**How to Reproduce**
The data paths are for GBO.

Load a file, select rows for IF 3 and scan 13.
```Python
from dysh.fits import GBTFITSLoad
sdf = GBTFITSLoad("/home/dysh/acceptance_testing/data/AGBT12A_076_05/AGBT12A_076_05.raw.acs")
mask = ( sdf["IFNUM"] == 3 ) & ( sdf["SCAN"] == 13 )
sdf["TIMESTAMP", "INTNUM"][mask]
```
that prints
```Python

TIMESTAMP | INTNUM
-- | --
2013_05_04_13:11:29 | 12
2013_05_04_13:11:29 | 12
2013_05_04_13:11:29 | 12
2013_05_04_13:11:29 | 12
2013_05_04_13:11:29 | 13
... | ...
2013_05_04_13:11:29 | 46
2013_05_04_13:11:29 | 47
2013_05_04_13:11:29 | 47
2013_05_04_13:11:29 | 47
2013_05_04_13:11:29 | 47

96 rows × 2 columns
```
For the same scan, but IF 0
```Python
mask = ( sdf["IFNUM"] == 0 ) & ( sdf["SCAN"] == 13 )
sdf["TIMESTAMP", "INTNUM"][mask]
```
prints
```Python

TIMESTAMP | INTNUM
-- | --
2013_05_04_13:11:29 | 0
2013_05_04_13:11:29 | 0
2013_05_04_13:11:29 | 0
2013_05_04_13:11:29 | 0
2013_05_04_13:11:29 | 1
... | ...
2013_05_04_13:11:29 | 34
2013_05_04_13:11:29 | 35
2013_05_04_13:11:29 | 35
2013_05_04_13:11:29 | 35
2013_05_04_13:11:29 | 35

96 rows × 2 columns
```

Notice that the first item has the same time stamp for both IFs, but the IF 3 starts with INTNUM 12 instead of 0 (like IF 0).

**Environment**
- Dysh version 0.4.0a
- Python version 3.9
- OS RH7

teuben:
Note this is the "nod9" example in my test_nodding.py - also reported in #376

mpound:
[...]

PR #392, opened 2024-09-26
Fix: group by scan and date-obs to generate integration numbers

astrofle:
This fixes issue #390 as is without breaking the tests.


src/dysh/fits/gbtfitsload.py
```diff
@@ -661,19 +661,15 @@ def _construct_integration_number(self):
...
+        dfs = self._index.groupby(["SCAN"])
+        for name, group in dfs:
+            dfst = group.groupby("DATE-OBS")
+            intnums = np.arange(0, len(dfst.groups))
+            for i, (n, g) in enumerate(dfst):
+                idx = g.index
+                intnumarray[idx] = intnums[i]
         self._index["INTNUM"] = intnumarray
 
         # Here need to add it as a new column in the BinTableHDU,
```
astrofle, line 675: @mpound are these lines needed?
I see you have a TODO below, but is that still the case?

mpound: @astrofle Yeah, the for loop around the @todo fix this MWP is still in need of fixing.  But given your changes, it would be done differently (and perhaps more easily) than I was trying to do it.

mpound (approved):
The use of pandas is very clever!

PR #395, opened 2024-10-02
Nodding2

teuben:
Although this PR needs dysh_data()  [see PR#394] this implements the getnod() function that is very similar to the getps() function.  The test_nodding.py in this PR contains 10 rich examples, of which only Example0 has been properly tested and compared with GBTIDL, since that dataset contains an FS, PS, and NOD scan on the same source. The jupyter notebook compares the PS and NOD for a user.

The test_nodding.py code (developed in spyder) contains the GBTIDL code used to generate the matching fits files.

A number of examples (see the top of the test_nodding.py code) do work, but there are issues to be resolved:

1.  calibration (VANE) missing for examples 1 and 3
2.  multiple feeds/scan uses in examples 6 and 9 (will need different masking)
3.  some examples don't seem to have well defined signal (examples 4,7,8)
4.  deprecated until revived and needed (examples 2 and 5)


teuben:
since there is a notebook check as well, and it still refers to the huge example (19GB) data, it will fail.

This notebook file need to be turned into a smaller file for the testdata tree.

astrofle:
What I've found so far:

- The scan class name should be `NodScan`, not `NODScan`, for consistency.
- When providing a single `scan` in `getnod`  the routine fails. It should be able to figure out which is the nod pair.
- The example notebook needs to be cleaned up: 
   * There are code cells that should not be there (e.g., cell with multiple definitions of `file1`).
   * There are markdown cells that provide incorrect information (e.g., they talk about doing polarization averaging, which is not done. They say that `getnod` will return two `ScanBlock`, but that should be one `ScanBlock` with two `NodScan`.

astrofle:
When the user passes an `fdnum` selection to `getnod`, `getnod` still reports the on-source and off-source beams. The data seems to be correct (`getnod` actually grabs the beams selected by `fdnum=`), but the message can be confusing.
For example:
```Python
from dysh.util.download import from_url
url = "http://www.gb.nrao.edu/dysh/example_data/nod-KFPA/data/TGBT22A_503_02.raw.vegas.trim.fits"
filename = from_url(url)

sdf = GBTFITSLoad(filename)

sp2 = sdf.getnod(scan=[62,63],ifnum=0,plnum=0,fdnum=[0,1]).timeaverage()

   Found nodding beams [2, 6]
```

astrofle:
One last thing that could be useful is to include a check for the user provided `fdnum`. Currently it will error out saying that the number of NodScans is odd, but it would be more informative to say something like: "`fdnum=xx` not present in the data", if the beam number is not present. 
I would also add `fdnum` to the docstring explicitly, since it has a different meaning than in other get methods (although the `**kwargs` docstring already explains how to use `fdnum` for `getnod`).

teuben:
I need to finish the notebook still. Other then the final version with figures, so it looks nice online, I wanted to do a baseline subtraction and compare the PS and NOD in (for now 2) diagrams. Baseline subtraction is still hampered because the inclusive option cannot handle units, so one needs 3 exclude regions, the edges and the line. Alternatively I could slice the spectra so it only has the baseline and the line, not the edges.

teuben:
notebook done, but figures not saved pending a final check by @astrofle  ?

astrofle:
> notebook done, but figures not saved pending a final check by @astrofle ?

The notebook looks good to me. Some comments below.

### Answers to the questions in the notebook:
* The KFPA has a roughly hexagonal beam layout around the central beam.
* Who is correct (nod vs ps)? If you look at the Nod using other beams, there is emission in them, suggesting that the source is more extended than the ~1.5' separation between beams. I'd take the `getps` answer as more accurate since the reference position is ~1 deg away from the target.
for example:
```Python
sp2 = sdf.getnod(scan=62,ifnum=0,plnum=0, fdnum=[3,0]).timeaverage()[1500:-1500]
sp2s = sp2.smooth('box',51)
sp2s.plot(title='dysh getnod',xaxis_unit="chan")
sp2s.stats()
```
### A few language corrections:
* after cell 7: "PS" observations -> "PS" observation
* after cell 7: we are dropped - we have dropped
* after cell 9: the 60 mK signal from the PS should have predicted more like 42 mK. -> the 60 mK signal from the PS observation should have been reduced to ~42 mK. (this could be partially due to the different system temperature of the beams, the Nod scan has a 18% higher system temperature.)

PR #400, opened 2024-10-10
Q7 align spectra

astrofle:
In this PR:
* Method to shift `Spectrum` given a number of channels.
* Refactor of the code used to shift. Now it can be found in `spectra.core`.
* Method to find the shift between two `Spectrum`.
* Method to align one `Spectrum` to another.
* Function to average spectra.
* Tests for all of the above.
* A notebook showing how to use the functions.


src/dysh/spectra/spectrum.py
```diff
@@ -1400,3 +1509,66 @@ def spectrum_reader_gbtidl(fileobj, **kwargs):
     # registry.register_writer("ipac", Spectrum, ascii_spectrum_reader_ipac)
     # registry.register_writer("votable", Spectrum, spectrum_reader_votable)
     # registry.register_writer("mrt", Spectrum, spectrum_reader_mrt)
+
+
+def average_spectra(spectra, equal_weights=False):
```
mpound, line 1514: This should be moved to core.py and eventually replace use of  average() in time averaging.

src/dysh/spectra/spectrum.py
```diff
@@ -1400,3 +1509,66 @@ def spectrum_reader_gbtidl(fileobj, **kwargs):
...
+    """
+    Average `spectra`.
+
+    Parameters
```
mpound, line 1518: Future enhancement: align=T/F to align all spectra to the first in the list.

src/dysh/spectra/spectrum.py
astrofle: I couldn't figure out how to check for `Spectrum` instances from core due to a circular import issue.

astrofle: Also, `core.average` and `spectrum.average_spectra` serve two different use cases. The former is for averaging inside a `Scan`, the later is for averaging `Spectrum`. To replace one with the other we would need to turn every integration into a `Spectrum`, which would be slower than needed.

src/dysh/spectra/spectrum.py
astrofle: Created issue #402

src/dysh/spectra/spectrum.py
```diff
@@ -818,6 +929,14 @@ def fake_spectrum(cls, nchan=1024, **kwargs):
         nchan : int, optional
             Number of channels. The default is 1024.
 
+        seed : {None, int, array_like[ints], SeedSequence, BitGenerator, Generator}, optional
```
mpound, line 932: These should be e.g., `~numpy.random.SeedSequence`  so the docs will link to the numpy docs/

mpound (approved):
Just fix that one doc issue in fake_spectra.  Very nice work!

Issue #252, opened 2024-05-12
baseline bug or feature -- debatable


leemundy:
**Describe the bug**
The baseline method has exclusion regions specified in channels. The baseline_model prints out the
solution with the slope in K/Hz rather than K/chan. The downside of the latter is that the y-intercept, c0, is not an obvious number since the spectra run over frequencies in the GHz range.

**How to Reproduce**
Use the provided frequency switch notebook. You will see that the spectral baseline is around zero K but the
baseline model is:
Model: Polynomial1D
Inputs: ('x',)
Outputs: ('y',)
Model set size: 1
Degree: 3
Parameters:
            c0                  c1           ...           c3          
            K                 K / Hz         ...        K / Hz3        
    ----------------- ---------------------- ... ----------------------
    7.735236225877181 -5.459874717654905e-09 ... -2.024830644563859e-63
It is not that it is incorrect; it just that the user does not have an intuitive check of
the value. It would be easier for the user to understand if the reported fit were in 
channel space. Also, all of the coefficients should be printed (c2 is missing in this
case).

**Environment**
- Dysh version
- Python version 3.11

teuben:
See also discussion in #174

teuben:
For other reasons (the rounding problem) the code has been fixed as to normalize the freq. axis to 0..nchan-1 so now the polynomial coeffs are more natural if you think in channel space.
Sadly, due to astropy not know what channels are, the units of the coefficients are still displayed as C1/Hz and C2/Hz2 etc., but they really are per channel

Issue #174, opened 2024-01-02
poorly constrained fit in example PS observation


teuben:
**Describe the bug**

Executing the example positionswitch I get this warning when a baseline fit is done

      WARNING: The fit may be poorly conditioned     [astropy.modeling.fitting]

Perhaps the freq axis was not "normalized", it claims it runs in the 1,400,000,000 band, so if the units used are Hz, I could imagine this error. Probably better to do the fit in km/s space, or even better, pick a center freq (or km/s) and fit around that such that the matrix is better behaved?

**How to Reproduce**

The cell with ta[0].baseline  does a baseline fit, and gives this warning.   

**Environment**

- Dysh version 0.2.0
- Python version 3.11.4
- OS Kubuntu 22.04

mpound:
what's your astropy version?

teuben:
5.3.2

teuben:
when upgraded to 6.0.0 I still get the warning.

mpound:
I can reproduce this with 0.2.0b.  
What do you mean by not normalized?  The spectral_axis of the spectrum is in Hz :
 ta[0].spectral_axis
Out[76]: 
<SpectralAxis 
   (observer to target:
      radial_velocity=0.0 km / s
      redshift=0.0
    doppler_rest=1420405751.7 Hz
    doppler_convention=optical)
  [1.41426369e+09, 1.41426297e+09, 1.41426226e+09, ..., 1.39082833e+09,
   1.39082762e+09, 1.39082690e+09] Hz>

Fitting needs to work regardless of what exclude region(s) are chosen and what units the axis are in.

teuben:
if the fitting is done in straight Hz, the X axis will be around 1.4e9, so the matrix to be inverted will have huge numbers that are all very positive. Better would be to have numbers scaled, like doppler velocities, or simply the mean of the freq's subtracted, and added back later when needed..

teuben:
It was found rescaling the freq. axis works (e.g. by first freq)

Using model=chebyshev'  it also works, since they need a rescaled freq

The differences are quite huge, I put a model 2nd order that went from 0 to 4 at the edge, and the returned spectral still had almost half of that left.   A little experimentation showed that normalizing to 1e9, 1e8, 1e7 worked, but already by 1e6 the fitted spectrum had rounding problems. This is for order=2, and is likely worse for higher orders.

@todo This also implies we should at least check what happens if the freq. axis is not regular 

[...]

Issue #86, opened 2023-10-05
Polynomial1D model not fitting at all, and other models are not matching gbtidl


etsmit:
First issue: Near misses

Using 'legendre' baseline model option fits nearly the same baseline as gbtidl, with most deviation at the edges of the baseline. The attached plot shows there is a difference in at least the third order between Dysh legendre and gbtidl, shown by the red line. (This is example dataset AGBT17A_404_...)

![bline_comparison_2](https://github.com/GreenBankObservatory/dysh/assets/24992070/b02ccd1c-f241-468b-95d6-d663f5cd7986)

Comparing the coefficients themselves is apples to oranges since dysh is using a Legendre model while gbtidl is using some form of basic polynomial sum(m) a_m x^m. Using the Polynomial1D model from the dysh side results in the second issue.

Second Issue: Polynomial1D fits are not working

A 2nd or 3rd order polynomial fit was applied to the data below, respectively. The 2nd order fit seemed to just find an average (indeed, all coefficients after the first are 0) and the 3rd order fit applied a straight, sloped line. Trying higher orders resulted in no change. The vertical black lines denote the regions of the fit.

![bline_comparison3](https://github.com/GreenBankObservatory/dysh/assets/24992070/012eb9df-304d-4760-ab55-8d5fc6c5f079)

![bline_comparison_5](https://github.com/GreenBankObservatory/dysh/assets/24992070/8dcb9def-522b-4615-b8bf-ba8a4da00dcd)

etsmit:
GBTIDL fitting code: /home/gbtidl/2.10.1/gbtidl/pro/toolbox/ortho_fit.pro

`;+
; Function uses general orthogonal polynomial to do least squares
; fitting.
;
; <p>This code came from Tom Bania's GBT_IDL work.  Local
; modifications include:
; <UL>
; <LI> Documentation modified for use by idldoc
; <LI> Array syntax changed to [] and compile_opt idl2 used.
; <LI> Indententation used to improve readability.
; <LI> Unnecessary code removed.
; <LI> Some argument checks added.
; </UL>
;
; @param xx {in}{required} The x-values to use in the fit.
; @param yy {in}{required} The data to be fit at xx.
; @param nfit {in}{required}{type=integer} The order of the polynomial to fit.
; @param cfit {out}{required} On return, cfit has coefficients of 
; the polynomial 
; <pre>
; fit = sum(m) a_m x^m 
; </pre> 
[...]

PR #403, opened 2024-10-15
86 baseline corrections

etsmit:
Resolves [Issue 86](https://github.com/GreenBankObservatory/dysh/issues/86).
Adds unit tests for baseline fitting.

Unit test for polynomial model baseline fitting commented out until issues [174](https://github.com/GreenBankObservatory/dysh/issues/174) and [252](https://github.com/GreenBankObservatory/dysh/issues/252) are resolved. (Unless they are?)


etsmit:
Also, raises python version to 3.10 and specutils to 1.17. I definitely messed that up.

astrofle:
Thanks a lot for doing this Evan.
The checks that are failing are only those for Python3.9, which makes sense given the bump in Python version. Let's talk about this today.

mpound (approved):

PR #406, opened 2024-10-16
Drop support for Python 3.9

tchamberlin:
[specutils has dropped support for Python 3.9](https://github.com/astropy/specutils/pull/1176). So, we will do the same

Note that per this spec, https://scientific-python.org/specs/spec-0000/ , we should drop support for Python 3.10 as of Q4 2024 (now). But I don't see any good reason for doing that, as it isn't really adding any maintenance burden and I haven't heard anyone clamoring for new 3.10+ specific features yet


tchamberlin:
It does not appear possible to use Python 3.13 using our maximum supported Astropy version. I don't think it's worth trying to understand why; we have no mandate to support 3.13 yet. So instead I have explicitly dropped support for Python 3.13+ in `pyproject.toml`

I have also regenerated our lock file (`requirements.txt`). But no Python dependency pins have changed

pyproject.toml
```diff
@@ -95,7 +91,7 @@ cov = "pytest --cov-report=xml --cov-config=pyproject.toml --cov=src/dysh --cov=
 
 # Run tests across all supported version of Python
 [[tool.hatch.envs.test.matrix]]
-python = ["39", "310", "311", "312"]
+python = ["310", "311", "312", "313"]
```
mpound, line 94: Should 313 be removed here?

tchamberlin: I think something got reverted during the rebase onto main. Will look tomorrow

mpound (approved):
merge conflicts need to be cleaned up but it looks good!

PR #407, opened 2024-10-16
Fix all(?) remaining invalid print->logger conversions

tchamberlin:
Found a few invalid log statements (via regex `\Wlogger\..*\(".*",`)


astrofle (approved):
Thanks!

PR #411, opened 2024-10-21
Change: align_to and shift now return a copy

astrofle:
This PR changes the behavior of `Spectrum.shift` and `Spectrum.align_to` to return a new `Spectrum`.

### DIFF
--- a/src/dysh/spectra/spectrum.py
+++ b/src/dysh/spectra/spectrum.py
@@ -542,23 +542,26 @@ class Spectrum(Spectrum1D, HistoricalBase):
             "fft" uses a phase shift.
         """
 
-        new_data = core.data_shift(self.data, s, remove_wrap=remove_wrap, fill_value=fill_value, method=method)
+        new_spec = self._copy()
+        new_data = core.data_shift(new_spec.data, s, remove_wrap=remove_wrap, fill_value=fill_value, method=method)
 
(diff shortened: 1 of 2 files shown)


src/dysh/spectra/spectrum.py
```diff
@@ -542,23 +542,26 @@ def shift(self, s, remove_wrap=True, fill_value=np.nan, method="fft"):
...
+        new_spec = self._copy()
+        new_data = core.data_shift(new_spec.data, s, remove_wrap=remove_wrap, fill_value=fill_value, method=method)
 
         # Update data values.
-        self._data = new_data
+        new_spec._data = new_data
 
         # Update metadata.
-        self.meta["CRPIX1"] += s
+        new_spec.meta["CRPIX1"] += s
```
mpound, line 552: Once we have flagging (masking) in place, we will also want to shift the mask?  But then have to decide what to do for non-integral channel shifts.

mpound (approved):

src/dysh/spectra/spectrum.py
astrofle: Yes, we will have to.

astrofle: we can test, maybe we'll have to round the fractional shifts.

PR #420, opened 2024-10-30
Issue 254

astrofle:
In this PR
* The following scans now ignore blanked integrations: `FSScan`, `PSScan`, `NodScan` (`TPScan` already had this, and `SubBeamNodScan` uses `TPScan`)
* Minor fix in PSScan where sigcalon data did not get the mask applied.


src/dysh/spectra/scan.py
```diff
@@ -977,11 +973,26 @@ def __init__(
...
+        self._refonrows = [self._refonrows[i] for i in goodrows]
+        self._refoffrows = [self._refoffrows[i] for i in goodrows]
+        self._sigcalon = self._sigcalon[goodrows]
+        self._sigcaloff = self._sigcaloff[goodrows]
+        self._sigonrows = [self._sigonrows[i] for i in goodrows]
+        self._sigoffrows = [self._sigoffrows[i] for i in goodrows]
+        # Update number of rows after removing blanks.
+        nsigrows = len(self._sigonrows) + len(self._sigoffrows)
+        self._nrows = nsigrows
+
```
mpound, line 994: lines 981-993 are repeated 3 times.   Add a todo to refactor?

src/dysh/spectra/core.py
```diff
@@ -109,12 +110,54 @@ def find_blanks(data):
...
+        nb3 = nb1
+    if cycle4 is not None:
+        nb4 = find_non_blanks(cycle4)
+    else:
+        nb4 = nb2
+    goodrows = reduce(np.intersect1d, (nb1, nb2, nb3, nb4))
+
+    if len(goodrows) != len(cycle1):
+        nblanks = len(cycle1) - len(goodrows)
+        logger.info(f"Ignoring {nblanks} blanked integration(s).")
```
mpound, line 156: I suggest the info message should be in the core routine rather than in the Scan classes.   Would there be other reasons to call find_non_blanks(), where we wouldn't need to notify the user? i.e. just cause you found them doesn't mean you are about to ignore them.

mpound (approved):
Very good.  Thanks for catching the missing mask, too!

PR #426, opened 2024-11-06
418 read/write of flags in GBTFITSLoad

mpound:
This PR adds 
* reading of GBTIDL flag files and creation of Flags object (sdf.flags) and data mask therefrom (sdf._sdf[i]._flagmask[k])
* the GBTFITSLoad constructor has a skipflags=T/F to control reading of GBTIDL flags files [default: False]
* Writing of flags to SDFITS via GBTFITSLoad.write(flags=True).   flags=True is the default.  A new column "FLAGS" with type unsigned int (1 or 0) and shape identical to the "DATA" column is added to the binary table HDU.
* new tests of FLAGS read/write and updates of old tests where necessary.  Tests include both multifile=True and False and flags=True and False.

This PR does NOT add:
* writing of flags from Scans for SDFITS output
* writing of flags directly from SDFITSLoad.  Although there is a flags argument in SDFITSLoad.write, it is currently ignored. In our case the FLAGS column is added before write() is called, so we don't want to double-add it.   We can think about how important it is to implement also in SDFITSLoad.


astrofle:
The reading of GBTIDL flags will fail if the flag was created after the fact through GBTIDL as it adds a space to the line, so the first character becomes `' *'` rather than `'*'` (fixed adding `v.strip()`).

It is also possible to create flags that include ranges so the flag file could look like:
```
[header]
created = Fri Nov 22 19:04:55 2024                                                                                                                                                                               
version = 1.0                                                                                                                                                                                                    
created_by = gbtidl                                                                                                                                                                                              
[flags]
#RECNUM,SCAN,INTNUM,PLNUM,IFNUM,FDNUM,BCHAN,ECHAN,IDSTRING
 *|20|42:51|*|*|*|2299|*|unspecified
```
This also implies that the flag only contains a start channel (BCHAN) but no end channel, which is currently not supported by the reader.

astrofle:
Not strictly related to this PR, but while trying to write a how-to guide I ran into these issues:
#429 
#428 
#427

astrofle:
Thanks a lot for all the work!
The only outstanding issue flag related is #432 (that is documented).
Let me know how you'd like to proceed.

mpound:
All three issues are now fixed.  I identified one more  during this testing.  I will make an issue but it should not hold up this merge.

astrofle (approved):

PR #430, opened 2024-11-22
Fix: support for flag files written by GBTIDL

astrofle:
Minor tweaks to support flag files written/modified by GBTIDL:
* Support for ranges (e.g., 41:52)
* Support for start channel with no end channel (does not work due to issues with how the flags are handled downstream, but the channel ranges are parsed "correctly")
* Support for end channel without start channel (does not work due to issues with how the flags are handled downstream, but the channel ranges are parsed "correctly")


astrofle:
I should write tests for this

PR #431, opened 2024-11-22
Fix: support for flag files written by GBTIDL

astrofle:
Minor tweaks to support flag files written/modified by GBTIDL:
* Support for ranges (e.g., 41:52)
* Support for start channel with no end channel
* Support for end channel without start channel
* Tests for the above
* Example recipe notebook for flagging


astrofle:
Draft until I get some examples written.

mpound (approved):
I made some minor (non-code) changes.   Thanks for catching this.

astrofle:
> To clarify my understanding of what GBTIDL can write:
> 
> 1. Only bchan/echan are allowed to have upper/lower limit by specifying * for one of them.
> 2. If there is a colon, it must separate two numbers.  You can't have e.g., "10:" to specify a lower limit.
> 
> Is that correct?

From what I've been able to find:
1. Yes.
2. Yes.

Issue #336, opened 2024-08-02
Notebook download outputs have white background in dark mode


vcatlett:
**Describe the bug**
In the notebooks rendered in the docs, the output from cells that download data via `dysh.util.download.from_url` have a white background in dark mode. 

**How to Reproduce**
See this page in dark mode: https://dysh.readthedocs.io/en/cat-docs-devel/examples/frequencyswitch.html#first-we-download-the-example-sdfits-data-if-necessary
![image](https://github.com/user-attachments/assets/5ac07a1e-c39b-470a-b278-3934b9cd8011)

astrofle:
This is the default behavior for widgets. See for example: https://myst-nb.readthedocs.io/en/latest/render/interactive.html

PR #433, opened 2024-12-02
Issue 336

astrofle:
In this PR:

- Changes to custom.css to use the same background color on all cells (so far). Fixes issue #336. 
- Updates to flagging notebook to show how to use `GBTFITSLoad.clear_flags` method.
- Removes transition (horizontal rule) at beginning of notebooks.
- Fixes to some of the docstrings that caused parsing errors.
- Add use of logger to `dysh.util.download`.
- Add sphinx-copybutton so that code cells can be copied with a click.

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -10,6 +10,20 @@ are more visible */
   background-color: #C6B9D2;
 }
 
+/*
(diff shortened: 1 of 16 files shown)


mpound (approved):
Looks good.

Issue #417, opened 2024-10-25
Spectrum.smooth does not preserve velocity frame and doppler convention. 


mpound:
**Describe the bug**
When you smooth a Spectrum, the resulting Spectrum does not inherit the velocity_frame, doppler_convention, observer, or _observer_location of the original spectrum.

**How to Reproduce**
1. velocity_frame, doppler_convention,  and observer
```
from dysh.fits import GBTFITSLoad
from dysh.util import get_project_testdata
f = get_project_testdata()/'TGBT21A_501_11/TGBT21A_501_11_scan_152_ifnum_0_plnum_0.fits'
g = GBTFITSLoad(f)
s = g.getspec(1)
print(f"{s.velocity_frame=}\n{s.doppler_convention=}\n{s.observer=}\n{s._observer_location=}")
```
s.velocity_frame='itrs'
s.doppler_convention='optical'
s.observer=<ITRS Coordinate (obstime=2021-02-10T07:38:37.500, location=(0., 0., 0.) km): (x, y, z) in m
    (882593.9465029, -4924896.36541728, 3943748.74743984)
 (v_x, v_y, v_z) in km / s
    (0., 0., 0.)>
s._observer_location=None
```
t = s.smooth("gaussian",16)
print(f"{t.velocity_frame=}\n{t.doppler_convention=}\n{t.observer=}\n{t._observer_location=}")
```
t.velocity_frame='fk5'
t.doppler_convention='optical'
t.observer=None
t._observer_location=None

2. observer_location
```
from dysh.spectra import Spectrum
from dysh.coordinates import Observatory
# use Spectrum `s` from example above.
s2 =  Spectrum(s.flux,observer_location=Observatory["ALMA"],meta=s.meta)
print(f"{s2._observer_location=}")
```
<EarthLocation (882593.9465029, -4924896.36541728, 3943748.74743984) m>
```
t2 = s2.smooth('gaussian',16)
print(f"{t2._observer_location=}") 
```
t2._observer_location=None  

**Environment**
- Dysh version 0.4.x
- Python version 3.10
- OS Ubuntu 22.04

astrofle:
@mpound I'm a bit conflicted about this. The two constructors for `Spectrum` can result in different behaviors which are not consistent. `Spectrum.make_spectrum()` will not add a `observer_location` attribute as is. Should we add that option or remove `observer_location` from `Spectrum.__init__`?

astrofle:
Some notes:
[...]

PR #436, opened 2024-12-05
Issue 417

astrofle:
In this PR:
- Solution for Issue #417.
- Test for Issue #417 .
- Remove `Spectrum._observer_location` (this breaks `Spectrum(s.flux,observer_location=Observatory["ALMA"],meta=s.meta)`).
- Adds `average` method for `Spectrum` objects to avoid having to import a separate function for averaging multiple spectra.


mpound (approved):
Nice job.  Note that PR #444 is now merged so you will may have some duplication to sort out.

Issue #372, opened 2024-09-03
Plotting changes `spectral_axis` of a `Spectrum`


astrofle:
**Describe the bug**
Plotting changes the underlying `Spectrum` object. This was unexpected.

**How to Reproduce**
```Python
from dysh.spectra import Spectrum
s = Spectrum.fake_spectrum()
ss = s._copy()
ss.plot(xaxis_unit="km/s", vel_frame="icrs")
# Now s and ss are different.
```

mpound:
ah, spectrum.velocity_axis_to should call with_frame() not set_frame().    that's my guess.

PR #437, opened 2024-12-06
401 changing the frame of a spectrum to itself changes its spectral axis

mpound:
The problem was that the output Spectrum observer attribute was not being set after the transform.  This is now fixed.  Also the way that topographic/ITRS coordinates were being handled was incorrect. Note these are the default coordinates for dysh spectra.  Finally, VELDEF needed to be changed on output to be consistent with velocity frame and convention.


mpound:
Currently failing because specplot.plot(vel_frame="topo") no longer works.   [notebook test]

mpound:
The fix, when found, was straightforward (make sure spectrum.observer == spectrum.spectral_axis.observer), but the effects rippled through other code.   In the process, I also discovered and fixed issue #372 .  New regression tests added for both this bug and 372.

PR #443, opened 2025-01-04
440 implement aperture efficiency calculations

mpound:
This PR includes the calculation and testing of aperture efficiency, surface error, gain correction as function of zenith distance.   New class `BaseGainCorrection`  is an abstract base class to be used by telescope specific gain correction classes, `GBTGainCorrection` is the GBT-specific class.   Important coefficients and constants are stored in an ASCII configuration file.   Test code is in test_gaincorrection.py

This PR also includes a new dysh/conf directory where configuration tables are to be stored. (The first is the gain correction table).  util code gets the conf path: `get_project_configuration() `

Not included in this PR is any change to the calibration routines to use these classes.


astrofle (approved):
Looks good. Thanks Marc.

PR #445, opened 2025-01-08
Q7 online

teuben:
This PR implements the equivalent of the GBTIDL routines "online" and "offline".

1. sdf = GBTOnline()

   Monitors new data in $SDFITS_DATA, $DYSH_DATA/sdfits or /home/sdfits (in that order).
   Any of the "sdf" functions "summary()", "gettp()" etc.
   will check if new data have arrived and refresh the "sdf" object automatically
   (user is informed if that happens).

2. sdf = GBTOnline('foo')

   Monitor a specific project. Possibly an unusual mode, but this mode exists.

3. sdf = GBTOffline('foo')

   Load a specific project, and is thus functionally similar to GBTFITSLoad('foo')


src/dysh/fits/tests/test_gbtfitsload.py
```diff
@@ -671,6 +672,32 @@ def test_add_history_comments(self):
         assert "ran the test for history and comments" in sdf.history
         assert any("Project ID: AGBT18B_354_03" in substr for substr in sb.history)
 
+    def test_online(self):
+        f1 = util.get_project_testdata() / "TGBT21A_501_11/TGBT21A_501_11.raw.vegas.fits"
+        f2 = util.get_project_testdata() / "TGBT21A_501_11/TGBT21A_501_11_2.raw.vegas.fits"
+        #
+        sdfits = "/tmp/sdfits"
```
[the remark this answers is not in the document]
teuben: indeed, this also forced me to use shutil.copyfille() which should fix the windows failure.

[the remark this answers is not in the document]
teuben: seems to become a long saga now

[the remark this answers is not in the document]
teuben: I've submitted  issue #447 and disabled the 2nd test for Windows. Thus passing the tests :-)

mpound:
You should add GBTOnline and GBTOffline to dysh/fits/__init__.py   so they can be imported directly from dysh.fits as we do with GBTFITSLoad. 
Arguably, dysh/shell/shell.py should also be updated, but I suggest that be made a separate enhancement issue for @tchamberlin after this branch is merged.

src/dysh/fits/gbtfitsload.py
```diff
@@ -2579,3 +2582,147 @@ def __setitem__(self, items, values):
...
+
+    @log_call_to_history
+    def __init__(self, fileobj=None, **kwargs):
+        self._online = fileobj
+        self._platform = platform.system()  # cannot update in "Windows":
+        # print("GBTOnline not supported on Windows yet, see issue #447")
+        if fileobj is not None:
+            self._online_mode = 1  # monitor this file
+            if os.path.isdir(fileobj):
+                GBTFITSLoad.__init__(self, fileobj)
```
mpound, line 2629: You need to pass on the other args such as skipflags.

src/dysh/fits/gbtfitsload.py
```diff
@@ -2579,3 +2582,147 @@ def __setitem__(self, items, values):
...
+            else:
+                self._online = dysh_data(fileobj)
+                GBTFITSLoad.__init__(self, self._online)
```
mpound, line 2632: Same here, pass on all *args

src/dysh/fits/gbtfitsload.py
```diff
@@ -2579,3 +2582,147 @@ def __setitem__(self, items, values):
...
+
+    Also note in GBTIDL one can use SDFITS_DATA instead of DYSH_DATA
+
+    """
+
+    @log_call_to_history
+    def __init__(self, fileobj, **kwargs):
+        self._offline = fileobj
+        self._filename = dysh_data(fileobj)
+        GBTFITSLoad.__init__(self, self._filename)
```
mpound, line 2602: Need *args in __init call.

src/dysh/fits/gbtfitsload.py
```diff
@@ -2579,3 +2582,147 @@ def __setitem__(self, items, values):
...
+        self._reload()
+        return super().getps(**kwargs)
+
+    def getnod(self, **kwargs):
+        self._reload()
+        return super().getnod(**kwargs)
+
+    def getfs(self, **kwargs):
+        self._reload()
+        return super().getfs(**kwargs)
```
mpound, line 2728: Need a catcher for subbeamnod

mpound (changes requested):
I tried it out on gb machines, works as expected except my comments about *args in GBTOnline/Offline.   Fix those and I'll approve.

src/dysh/fits/gbtfitsload.py
teuben: plus the maybe different name later on

teuben:
Can't seem to write the __init__

mpound:
> reminding myself too: can't write the dunder **init**, but have to write with two backslashes up front __init__ and the rest is history

gets me every time

mpound (approved):
Nice job!   I tried on on gb machine, works as expected.

teuben:
Need to keep an eye open now for performance (e.g. is there a longer and longer wait) as the file grows

PR #448, opened 2025-01-16
Issue 174

astrofle:
This PR fixes issue #174 by using the domain parameter of the `astropy.modeling.polynomial` models. However, this solution does not fully work for `Polynomial1D`, since when using `domain != (-1,1)` you cannot evaluate a `Polynomial1D` if it has units (or at least I couldn't figure out how). To solve this, I chose to remove the units from the `Polynomial1D` model using the functions provided by `specutils`. This has the downside of losing some information, like the uncertainties in the estimated parameters -- which are already lost for the other models since they go through this unit stripping.

### DIFF
--- a/src/dysh/spectra/core.py
+++ b/src/dysh/spectra/core.py
@@ -19,6 +19,8 @@ from astropy.modeling.polynomial import Chebyshev1D, Hermite1D, Legendre1D, Poly
 from scipy import ndimage
 from specutils import SpectralRegion
 from specutils.fitting import fit_continuum
+from specutils.fitting.fitmodels import _strip_units_from_model
+from specutils.utils import QuantityModel
 
 from ..coordinates import veltofreq
 from ..log import log_function_call, logger
@@ -360,7 +362,9 @@ def baseline(spectrum, order, exclude=None, exclude_region_upper_bounds=True, **
     model = minimum_string_match(kwargs_opts["model"], list(available_models.keys()))
     if model == None:
         raise ValueError(f'Unrecognized input model {kwargs["model"]}. Must be one of {list(available_models.keys())}')
-    selected_model = available_models[model](degree=order)
+    sa_min = spectrum.spectral_axis.min().value
+    sa_max = spectrum.spectral_axis.max().value
+    selected_model = available_models[model](degree=order, domain=(sa_max, sa_min))
 
     _valid_exclude_actions = ["replace", "append", None]
     if kwargs_opts["exclude_action"] not in _valid_exclude_actions:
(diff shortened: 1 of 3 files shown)


src/dysh/spectra/spectrum.py
```diff
@@ -241,6 +241,11 @@ def baseline(self, degree, exclude=None, include=None, **kwargs):
         }
         kwargs_opts.update(kwargs)
 
+        # Switch to GHz to avoid poorly conditioned fit warnings due to large x-axis values.
+        # See Issue 174.
+        org_unit = self._spectral_axis.unit
+        self._spectral_axis = self._spectral_axis.to("GHz")
```
mpound, line 247: This won't work if using a spectrum where the spectral axis cannot be converted to frequency units. e.g. after reading in a GBTIDL spectrum that was written in km/s. See e.g. testdata/gbtidl_spectra/onoff-L_getps_152_OPTI-HEL.ascii.gz.    Or if the spectral axis is in channel space.   Perhaps add an if statement so that the change to GHz only occurs if the spectral_axis is frequency.  Also, the normalize keyword that Peter added was a different fix to this bug.   We likely don't need it if we use your fix.

astrofle: Agree. I tried a new solution. It also has drawbacks, mainly because of a possible bug with `astropy` (or my lack of understanding on how to use their `Polynomial1D` model with a domain).

astrofle: I added a test for the case of a `Spectrum` with velocity units in its `spectral_axis`. Good idea.

src/dysh/spectra/core.py
```diff
@@ -539,15 +543,29 @@ def baseline(spectrum, order, exclude=None, exclude_region_upper_bounds=True, **
...
 
+    # astropy's Polynomial1D model does not work well when using
+    # a domain other than (-1,1), so we need to remove the units
+    # of the model. We use `specutils` methods to do so. This has the
+    # downside that the uncertainties of the model are removed.
+    if not isinstance(fitted_model, QuantityModel):
+        strip_model, _, _ = _strip_units_from_model(fitted_model, p)
+        strip_model.domain = fitted_model.domain
+        strip_model.window = fitted_model.window
+        fitted_model = QuantityModel(strip_model, p.spectral_axis.unit, p.flux.unit)
```
mpound, line 565: I like this. We are always ensured of returning the same type from this method call.   It helps for another method I want to add (post-release).

src/dysh/spectra/spectrum.py
```diff
@@ -172,82 +172,49 @@ def baseline(self, degree, exclude=None, include=None, **kwargs):
...
-                Default: False
+            Default: no exclude region
 
+        include: list of 2-tuples of int or `~astropy.units.Quantity`, or `~specutils.SpectralRegion`
+            List of region(s) to include in the fit.  The tuple(s) represent a range in the form [lower,upper], inclusive.
+            See `exclude` for examples.
+        model : str
+            One of 'polynomial' 'chebyshev', 'legendre', or 'hermite'
+            Default: 'chebyshev'
+        fitter  :  `~astropy.fitting._FitterMeta`
```
mpound, line 198: ~astropy.modeling.fitting._FitterMeta 

(this link has always been wrong)

src/dysh/spectra/spectrum.py
```diff
@@ -172,82 +172,49 @@ def baseline(self, degree, exclude=None, include=None, **kwargs):
...
+            The fitter to use. Default: `~astropy.fitter.LinearLSQFitter` (with `calc_uncertaintes=True`).
```
mpound, line 199: ~astropy.modeling.fitter.LinearLSQFitter

src/dysh/spectra/spectrum.py
mpound: also should be corrected in core.baseline.

mpound (approved):
Excellent. Thanks for doing this. 
A few minor doc fixes is all.

Issue #260, opened 2024-05-23
baseline exclude= specifications complain if units are used


teuben:
**Describe the bug**
The example for baseline subtraction claim we can use units in the exclude=, but it fails .

**How to Reproduce**

In the FS notebook I used

      exclude = [(-1500*kms,-400*kms),(-200*kms,50*kms),(600*kms,2200*kms)]

which caused the error

      ...
      File ~/GBT/dysh/src/dysh/spectra/core.py:126 in exclude_to_region
         if len(np.shape(exclude[0])) == 0:

       File ~/GBT/anaconda3/lib/python3.11/site-packages/numpy/core/fromnumeric.py:2024 in shape
         result = asarray(a).shape

       File ~/GBT/anaconda3/lib/python3.11/site-packages/astropy/units/quantity.py:1352 in __float__
         raise TypeError(

     TypeError: only dimensionless scalar quantities can be converted to Python scalars

**Environment**
- Dysh version
- Python version
- OS

astrofle:
Just to be pedantic, the example notebook says explicitly: `Right now, compound exclusions only work in channel space.`

PR #449, opened 2025-01-16
Issue 260

astrofle:
This PR expands the use of exclusion/inclusion regions to accept multiple ranges with units or not.

Once finished this will fix issue #260.

This also adds tests for baseline fitting using `include`, and using regions with units. The tests with units do not check for "quality", just that the region selection does not error.


src/dysh/spectra/core.py
mpound, line 1: Change print on line 542 to logger.info

mpound (approved):
Excellent, very comprehensive!

PR #464, opened 2025-01-31
457 using flagging data in some dataset is broken

mpound:
This fixes unsorted indices in the Flag astropy Table, which was causing the error.


mpound:
All our Windows CI have been failing because of a pip cache issue.

astrofle:
> All our Windows CI have been failing because of a pip cache issue.

No idea why, but when rerunning the failed jobs they complete.

teuben (approved):
good sleuthing Marc.

PR #472, opened 2025-02-05
Beta testing 0.4.2

astrofle:
Should go to main as well.


mpound:
my plan is to merge release-0.4 into main once we have everything we want in the release. No need to do individual PRs to main.

PR #487, opened 2025-02-24
441 implement opacity correction

mpound:
This PR implement WeatherForecast classes to interface with the GBT weather script getForecastValues.  It also adds atmospheric temperature and zenith opacity methods to GainCorrection classes to use these.   There is test code that uses a `testmode=True` keyword that will read ASCII files in testdata/ that were output by getForecastValues in order to test as much of the new code as possible.  

Also include documentation fixes and an new utility method for creating MJD lists.


astrofle:
I tried the following:
```Python
from dysh.util.weatherforecast import GBTForecastScriptInterface
g = GBTForecastScriptInterface()
g(vartype="Opacity", mjd=60733.2916667, freq=50, coeffs=False)
```

I was expecting a single value, but instead I got values for the default range of frequencies and the current time. It seems the `__call__` method is not passing the arguments when `coeff=False`.

mpound:
Argument list in both cases is not correctly handled.   Note that in general a user wouldn't use `GBTForecastScriptInterface` directly, but rather through `GBTWeatherForecast` which can handle parameters of differing types like Quantity and Type, whereas  `GBTForecastScriptInterface` requires lists of native types.  However your direct test is important!  So your example should be:
```
from dysh.util.weatherforecast import GBTForecastScriptInterface
g = GBTForecastScriptInterface()
g(vartype="Opacity", mjd=[60733.2916667], freq=[50], coeffs=False)
```

This correctly returns the same as ` /users/rmaddale/bin/getForecastValues -timeList 60733.2916667 -freqList  50 -type Opacity` 
```
# Opacity(50,60733.2916667) = 0.3125
array([6.07332917e+04, 5.00000000e+01, 3.12500000e-01])
```

It would be good to have a `testmode=True` case for this -- make it an issue?

mpound:
Added check for lower limit on input date.

This PR satisfies 2.5.8.3-R1 , 2.5.8.3-R2, and 2.5.8.3-R3

astrofle:
The following two ways of calling `GBTWeatherForecast.fetch` fail. I don't know if this should be fixed in the method signature, or if the method should make `specval` an array before calling `GBTForecastScriptInterface`. 

```Python
from astropy import units as u
from dysh.util.weatherforecast import GBTWeatherForecast

gw = GBTWeatherForecast()

gw.fetch(vartype="Opacity", mjd=60733.2916667, specval=50*u.GHz, coeffs=False)
gw.fetch(vartype="Opacity", mjd=60733.2916667, specval=[50*u.GHz], coeffs=False)
```

astrofle:
This could be improved by catching bad values an returning something that would raise an error/warning:
```Python
from dysh.util.weatherforecast import GBTWeatherForecast

gw = GBTWeatherForecast()
f = Spectrum.fake_spectrum()
gw.fetch(vartype="Opacity", mjd=60733.2916667, specval=f.spectral_axis, coeffs=False)
```
Now it returns -9999.0. If one is not paying attention I'm guessing that would blow when trying to correct for the opacity.

astrofle:
```Python
from dysh.util.gaincorrection import GBTGainCorrection
f = Spectrum.fake_spectrum()
ggc = GBTGainCorrection()
ggc.zenith_opacity(specval=f.spectral_axis)
```
produces an error because it uses `type` instear of `vartype` on the call to `get_weather`

astrofle:
```Python
from dysh.util.gaincorrection import GBTGainCorrection
f = Spectrum.fake_spectrum()
ggc = GBTGainCorrection()
ggc.get_weather(vartype="Opacity", specval=f.spectral_axis)
```
Results in:
```Python
AttributeError: 'GBTGainCorrection' object has no attribute '_forecast'
```

astrofle:
```Python
from dysh.util.weatherforecast import GBTWeatherForecast

gw = GBTWeatherForecast()
f = Spectrum.fake_spectrum()
gw.fetch(vartype="Opacity", mjd=f.obstime, specval=f.spectral_axis, coeffs=False)
```
results in 
```Python
File [/home/sandboxes/psalas/Dysh/dysh/src/dysh/util/weatherforecast.py:57](http://localhost:9050/home/sandboxes/psalas/Dysh/dysh/src/dysh/util/weatherforecast.py#line=56), in GBTWeatherForecast.fetch(self, specval, vartype, mjd, coeffs)
     55 mjd_list = to_mjd_list(mjd)
     56 if mjd_list is not None:
---> 57     bad_date = [x < self.LOWER_MJD_LIMIT for x in mjd_list]
     58     if any(bad_date):
     59         raise ValueError("There is no weather date before MJD 53129[/2004-May-05.](http://localhost:9050/2004-May-05.)")

TypeError: 'numpy.float64' object is not iterable
```

mpound:
> The following two ways of calling `GBTWeatherForecast.fetch` fail. I don't know if this should be fixed in the method signature, or if the method should make `specval` an array before calling `GBTForecastScriptInterface`.
> 
> ```python
> from astropy import units as u
> from dysh.util.weatherforecast import GBTWeatherForecast
> 
> gw = GBTWeatherForecast()
> 
> gw.fetch(vartype="Opacity", mjd=60733.2916667, specval=50*u.GHz, coeffs=False)
> gw.fetch(vartype="Opacity", mjd=60733.2916667, specval=[50*u.GHz], coeffs=False)
> ```

I feel the first should be made to work, but not the second.  50*u.GHz is a Quantity so matches the signature.  However, [50*u.GHz] is a list which does not.  However, the latter may be a common mistake so perhaps we should support it. Ugh.

mpound:
> ```python
> ggc.get_weather(vartype="Opacity", specval=f.spectral_axis)
> ```

Oooh, you've also uncovered the case of passing in a SpectralAxis which is derived from Quantity so needs to be handled correctly.

mpound:
> ```python
> from dysh.spectra import Spectrum
> from dysh.util.weatherforecast import GBTWeatherForecast
> 
> gw = GBTWeatherForecast()
> f = Spectrum.fake_spectrum()
> gw.fetch(vartype="Opacity", mjd=f.obstime, specval=f.spectral_axis, coeffs=False)
> ```
> 
> results in
> 
> ```python
> File [/home/sandboxes/psalas/Dysh/dysh/src/dysh/util/weatherforecast.py:57](http://localhost:9050/home/sandboxes/psalas/Dysh/dysh/src/dysh/util/weatherforecast.py#line=56), in GBTWeatherForecast.fetch(self, specval, vartype, mjd, coeffs)
>      55 mjd_list = to_mjd_list(mjd)
>      56 if mjd_list is not None:
> ---> 57     bad_date = [x < self.LOWER_MJD_LIMIT for x in mjd_list]
>      58     if any(bad_date):
>      59         raise ValueError("There is no weather date before MJD 53129[/2004-May-05.](http://localhost:9050/2004-May-05.)")
> 
> TypeError: 'numpy.float64' object is not iterable
> ```

This is because of issue #491.  Fixing that fixes this. 

> If we pass `mjd` as a list:
> 
> ```python
> gw.fetch(vartype="Opacity", mjd=[f.obstime], specval=f.spectral_axis, coeffs=False)
> ```
> 
> Then:
> 
> ```python
> 55 mjd_list = to_mjd_list(mjd)
>      56 if mjd_list is not None:
> ---> 57     bad_date = [x < self.LOWER_MJD_LIMIT for x in mjd_list]
>      58     if any(bad_date):
>      59         raise ValueError("There is no weather date before MJD 53129[/2004-May-05.](http://localhost:9050/2004-May-05.)")
> 
> TypeError: '<' not supported between instances of 'Time' and 'int'
> ```

This example is incorrect use of  the `Time` class, which can hold multiple values and acts as its own list.  So mjd=[Time object]  is wrong.  I will address this in documentation for `fetch()` which I notice has none.

mpound:
Here's an interesting question -- what do we do for frequencies <2 GHz?  The coefficient table minimum frequency is 2 GHz, so no interpolation is possible if `coeffs=True` and an exception is raised because this case is not currently handled.  For `coeffs=False` the returned value(s) are -9999 which is what `getForecastValues` returns.    I think in general I will issue a warning is any(values) == -9999 no matter what `vartype` was asked for.

astrofle:
Per conversation on 2025-03-03: "If a user requests the opacity for a frequency below 2 GHz, return the opacity at 2 GHz."

mpound:
* Frequencies < 2GHz now return values for 2GHz.  

* Other issues
There is now a script util/test/test_weather_gbo_only.py that tests all the issues found in reviewing this PR.  If other issues are found (and subsequently fixed), they should be added to that test script.   You can run it with 
  ```
  pytest /test_weather_gbo_only.py
  ```
  or just
  ```
  ./test_weather_gbo_only.py
  ```

   It will be ignored in regular pytest (see pyproject.toml)

* `to_mjd_list` and `to_quantity_list` added to allow user convenience on inputs

astrofle:
I added the option of not passing a frequency or `specval` since some `vartype` do not require it (e.g., `"Winds"`).

astrofle (approved):
Great work! Thanks a lot Marc.

Issue #494, opened 2025-03-04
Column names not recognized


swapnaneeldey:
Column names not recognized in GBTFITSLoad after deleting flags using GBTIDL? 

To reproduce this error, 
- Take any raw Vegas files with flags. Open it in GBTIDL and then delete flags using the unflag command in the terminal. 
- If you now load the file using GBTFITSLoad, it gives the following error: 
 Exception: Column names ['*|6|*|*|*|*|22540|22540|unspecified'] do not match expectated ['RECNUM', 'SCAN', 'INTNUM', 'PLNUM', 'IFNUM', 'FDNUM', 'BCHAN', 'ECHAN', 'IDSTRING']

**Environment**
- Dysh version
- Python version
- OS

astrofle:
Thanks a lot for reporting this bug. It has now been fixed in the main branch. To install the patched version you can use:
```bash
python -m pip install "dysh[all] @ git+https://github.com/GreenBankObservatory/dysh"
```
that will install `dysh` with all the extra dependencies from the main branch of the GitHub repository. If you do not want all of the extra dependencies remove the `[all]`.

If updating `dysh` is not an option, you can "fix" the .flag files by removing any extra lines that start with a # after the column definition. For example, in the following file you'd remove the last line:
```
[header]
created = Fri Apr  5 13:52:30 2024
version = 1.0
created_by = sdfits
[flags]
#RECNUM,SCAN,INTNUM,PLNUM,IFNUM,FDNUM,BCHAN,ECHAN,IDSTRING
*|6|*|*|2|0|3072|3072|VEGAS_SPUR
#*|6|*|*|2|0|3072|3072|VEGAS_SPUR
```

Please let me know if this solves the issue or if you have additional questions.

PR #495, opened 2025-03-05
Issue 494

astrofle:
Fixes issue #494


mpound (approved):
Thanks, glad it was simple!

PR #502, opened 2025-03-10
489 implement tasnu scaling factors

mpound:
brightness temperature scaling


src/dysh/spectra/scan.py
```diff
@@ -210,11 +352,11 @@ def fdnum(self):
     @property
     def pols(self):
         """The polarization number(s)
-
-        Returns
-        -------
-        list
-            The list of integer polarization number(s)
+        gc.physical_aperture
```
astrofle, line 355: clean up

src/dysh/util/gaincorrection.py
```diff
@@ -276,16 +321,29 @@ def aperture_efficiency(
...
+        For mutiple dates:
+        - For one spectral value and one angle, the aperture efficiency at each date is returned.
+        - For multiple spectral values and multiple angles, it is assumed they go together, so the lengths must match
+          the number of dates. The aperture efficiency for each (spectral value, angle, date) tuple will be returned.
+
         Parameters
         ----------
-        specval : `~astropy.units.quantity.Quantity`
-            The spectral value -- frequency or wavelength -- at which to compute the efficiency
+        specval : `~astro.units.quantity.Quantity`
```
astrofle, line 339: astropy

mpound:
@astrofle I still have to add methods on ScanBlock to call the Scan methods, but there is enough for you to start reviewing now.

astrofle:
For a separate issue:
* the units are not reflected in the plots

astrofle:
The units are not propagated to `Spectrum` objects. The `flux` attribute is in K even if `bunit='jy'` is used, and `"TUNIT7"` is in "Counts" (I guess this was an issue before, since it should have been updated to K).
Example:
```Python
from dysh.fits import GBTFITSLoad
filename = get_project_testdata() / "AGBT05B_047_01/AGBT05B_047_01.raw.acs"
sdf = GBTFITSLoad(filename)
pssb_jy = sdf.getps(scan=51, ifnum=0, plnum=0, bunit="jy", zenith_opacity=0.08)
ps_jy = pssb_jy.timeaverage()
ps_jy.meta["TUNIT7"] == "Counts" # This is True, but shouldn't be.
ps_jy.flux.unit.to_string() == 'K' # This is also True, and shouldn't.
```

astrofle:
The argument of the exponential should be positive here (also in the comment above):
https://github.com/GreenBankObservatory/dysh/blob/eea079e5ea9cf31bc4c26ed6d225627d90d2e29b/src/dysh/util/gaincorrection.py#L422C1-L423C1

astrofle (approved):
Thanks Marc!

PR #509, opened 2025-03-13
specplot-test-fix

astrofle:
Fix: plot will not show when running test with a display

### DIFF
--- a/src/dysh/plot/tests/test_specplot.py
+++ b/src/dysh/plot/tests/test_specplot.py
@@ -1,6 +1,7 @@
 """Tests for specplot."""
 
 from pathlib import Path
+from unittest.mock import patch
 
 from dysh.fits import GBTFITSLoad
 from dysh.util import get_project_testdata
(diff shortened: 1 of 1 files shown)


src/dysh/plot/tests/test_specplot.py
```diff
@@ -9,7 +10,8 @@
 class TestSpecplot:
     """ """
 
-    def test_savefig(self, tmp_path):
+    @patch("dysh.plot.specplot.plt.show")
```
mpound, line 13: Isn't the attribute `specplot._plt` not `specplot.plt`? Or does this refer to the imported "matplotlib.pyplot as plt"?

astrofle: The later.

astrofle: However, I do not fully understand the details, but it works.

mpound (approved):

PR #513, opened 2025-03-14
489-implement-tasnu-scaling-factors-PS-review

astrofle:
Fix: minor changes to docstrings and make zenith opacity optional for 'ta' scaling

### DIFF
--- a/src/dysh/spectra/scan.py
+++ b/src/dysh/spectra/scan.py
@@ -141,8 +141,9 @@ class ScanBase(HistoricalBase, SpectralAverageMixin):
 
     @classmethod
     def _check_bunit(self, bunit):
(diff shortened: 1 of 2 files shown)


src/dysh/spectra/scan.py
```diff
@@ -259,8 +261,10 @@ def scale(self, bunit, zenith_opacity):
         s = bunit.lower()
         if s == self._bunit:
             return
+        if bunit != "ta" and zenith_opacity is None:
```
mpound, line 264: This should be s !="ta"  rather than bunit, since bunit is allowed to be any case.

astrofle: thanks

PR #514, opened 2025-03-18
Fix SyntaxWarning for invalid escape sequences

tchamberlin:
Same issue as #119

This can be checked for via `ruff check --select W605`


mpound (approved):
Thanks for catching these.

PR #516, opened 2025-03-18
Cap astropy<7

astrofle:
dysh does not work with `astropy>=7`. This should help avoid using astropy7 with dysh.


mpound:
If the only issue is that sdfitsload `update` calls need to change to `update_header`, then perhaps a better solution is to change those calls and make astropy>=6.1 be the lower bound.    So a mix of this PR and #497.

PR #517, opened 2025-03-19
No cal nod

astrofle:
This PR 
* adds support for no noise diode for `getnod`
* adds tests for `getnod` with and without noise diode


mpound (approved):
Looks good!  At some point we need to clean up testdata/README.md (or delete it) but that doesn't have to be part of this PR.

teuben:
note this PR has been merged via the vane PR

PR #518, opened 2025-03-19
Quadrant detector methods

astrofle:
This PR:
* `qd_check` checks that the pointing error registered by the quadrant detector (QD) is less than a user provided fraction of the beam width.
* `qd_flag` flags any rows where the QD pointing error is larger than the user provided fraction of the beam width.
* `qd_correct` applies the pointing corrections registered by the QD.


src/dysh/coordinates/core.py
```diff
@@ -669,3 +671,63 @@ def get_earth_location(longitude, latitude, height=0.0, ellipsoid=None):
...
+    Equatorial to horizontal coordinate conversion.
+
+    Parameters
+    ----------
+    lon : float
+        Longitude coordinate. E.g., RA or Galactic longitude.
+    lat : float
+        Latitude coordinate. E.g., Dec or Galactic latitude.
+    frame : str
+        Input coordinate frame. Must be recongized by `~astropy.coordinates`
```
mpound, line 687: typo - recognized

src/dysh/fits/gbtfitsload.py
```diff
@@ -2627,6 +2630,145 @@ def __setitem__(self, items, values):
                 " You may wish to update the selection. "
             )
 
+    @log_call_to_history
+    def qd_correct(self, ignore_jump: bool = False) -> None:
+        """
+        Apply quadrant detector (QD) corrections to antenna position.
```
mpound, line 2636: Can you add a reference link to what the quadrant detector is and why one might wish to apply corrections?  Is there GBT docs on this?   Also,  "corrections to antenna position" sounds like you are correcting the telescope location, whereas the code is updating the sky coordinates of sources.    That should be clarified.

mpound (approved):
Very nice additional functionality.  A few small comments.

src/dysh/coordinates/core.py
astrofle: good catch. Thanks!

Issue #512, opened 2025-03-14
Reported scale of Scans and Spectra can be incorrect.


mpound:
**Describe the bug**
There is inconsistency between header BUNIT, TUNIT7, and Spectrum.flux_unit. 

**How to Reproduce**
The units are not propagated to Spectrum objects. The flux attribute is in K even if bunit='jy' is used when scaling brightness temperature, and "TUNIT7" is in "Counts"
Example:

```
from dysh.fits import GBTFITSLoad
from dysh.util import get_project_testdata
filename = get_project_testdata() / "AGBT05B_047_01/AGBT05B_047_01.raw.acs"
sdf = GBTFITSLoad(filename)
pssb_jy = sdf.getps(scan=51, ifnum=0, plnum=0, bunit="jy", zenith_opacity=0.08)
ps_jy = pssb_jy.timeaverage()
ps_jy.meta["TUNIT7"] == "Counts" # This is True, but shouldn't be.
ps_jy.flux.unit.to_string() == 'K' # This is also True, and shouldn't.
The same is true if you take one from a Scan:

ps_jy_i = pssb_jy[0].calibrated(0)
ps_jy_i.meta["TUNIT7"] == "Counts" # True
ps_jy_i.flux.unit.to_string() == 'K' # True
```

**Environment**
- Dysh version 0.4
- Python version 3.11
- OS Ubuntu

mpound:
Question:  should meta["TUNIT7"] be changed if we have meta["BUNIT"] correct?   My thought is that TUNIT7 serves as a history of the units of the raw integration.

astrofle:
Looking at what Bob wrote for the `gbtgridder`, "TUNIT7" should also be updated (i.e., the `gbtgridder` uses "TUNIT7" to set "BUNIT" of the data cubes).

PR #530, opened 2025-03-20
Scanblock refactor

mpound:
This PR refactors the Scan classes, ScanBlock, and the calibration routines in GBTFITSLoad.   Nearly(?) all copypasta code is removed in factor of common methods.    Sub-issues are also addressed.  Notebooks have been updated to conform to the new requirement of fdnum, plnum, ifnum in calibration routines. 

Tutorial notebooks have *not* been updated.


astrofle:
Picking more than one `plnum` raises an exception, but selecting more than one `ifnum` or `fdnum` does not. Is that what we want? I would expect consistent behavior (all raise an exception, or none does). I think raising an exception should be the behavior.

When selecting more than one `ifnum` or `fdnum`, there is no warning that some of the values are being ignored.

mpound:
Yes, they should all raise exceptions.  I'll make the modification.

Technically a ScanBlock could include multiple plnum/ifnum etc if the user constructed it themselves for some reason.  Unlikely, but possible.   There should be a summary() for ScanBlock though to make it easy to see what's inside.  I will add as an RFE issue.

astrofle:
And yes to updating "TUNIT7" (see https://github.com/GreenBankObservatory/dysh/issues/512#issuecomment-2761774723).

mpound:
ValueError now raised in all cases.  Had to get rid of some old, now incorrect, code.

astrofle (approved):

PR #531, opened 2025-03-21
Config file

astrofle:
Adds the use of a configuration file. The configuration file is created when launching `dysh`. By default the configuration file goes into: `$XDG_CONFIG_HOME/.dysh/config`
There is an option to skip the creation: `--skip-config`. 
When launching the shell, there is a INFO message stating that the configuration file was created (see below).

Image showing the banner after launching the dysh shell
The location of the message could be improved.

### DIFF
--- a/src/dysh/config/core.py
+++ b/src/dysh/config/core.py
@@ -1,0 +1,227 @@
+"""Configuration core functions"""
+
+import contextlib
+import importlib
+import io
+import os
+import pkgutil
+import warnings
+from contextlib import nullcontext
+from pathlib import Path
+from textwrap import TextWrapper
+
+from astropy import config as ac
+from astropy.config import configuration as acc
+from astropy.utils import silence
+from astropy.utils.exceptions import AstropyDeprecationWarning
+
+__all__ = ["create_config_file", "get_config_dir_path"]
+
+
+# def create_config_file(package: str = "dysh", rootname: str = "dysh", overwrite: bool = False) -> bool:
+#    """
+#    Create the default configuration file for the specified `package`.
+#    If the file already exists, it is updated only if it has not been modified.
+#    Otherwise the `overwrite` flag is needed to overwrite it.
+#
+#    Parameters
+#    ----------
+#    package : str
+#        The package.
+#    rootname : str
+#        Name of the root configuration directory.
+#    overwrite : bool
+#        Force updating the file if it already exists.
+#
+#    Returns
+#    -------
+#    updated : bool
+#        If the profile was updated, `True`, otherwise `False`.
+#    """
+#
+#    return ac.create_config_file(package, rootname=rootname, overwrite=overwrite)
+
+
+def get_config_dir_path(rootname: str = "dysh") -> Path:
+    """
+    Determines the package configuration directory name and creates the
+    directory if it doesn't exist.
+
+    This directory is typically ``$HOME/.astropy/config``, but if the
+    XDG_CONFIG_HOME environment variable is set and the
+    ``$XDG_CONFIG_HOME/astropy`` directory exists, it will be that directory.
+    If neither exists, the former will be created and symlinked to the latter.
+
+    Parameters
+    ----------
+    rootname : str
+        Name of the root configuration directory. For example, if ``rootname =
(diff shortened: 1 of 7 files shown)


mpound:
This infrastructure is good to have and the initial implementation works.  I can add a Conf to plot/__init__.py and it gets written to ~/.dysh/config/dysh.cfg.  But I don't understand how the user uses it.  For instance, if I edit ~/.dysh/config/dysh.cfg, say to change summary_max_rows, it gets overwritten unless I use --skip-config.   However, even if I use --skip-config summary_max_rows gets reset to 1000 by fits `__init__.py`.    Furthermore, if I change it on the command line with `conf.summary_max_rows=2`, it has no effect because there is no active connection to `pd.set_options`.   Moving the line `pd.set_option("display.max_rows", conf.summary_max_rows)` to inside of `summary`  fixes this.

It would be good to have a utility code `dysh.show_config()` that would introspect all modules and print out their conf items if they have a conf object.   Not required for this PR, just an RFE.
 
It may take some time before we figure out how best to use this infrastructure and make it available to users.

src/dysh/fits/gbtfitsload.py
```diff
@@ -39,12 +39,16 @@
...
 from ..util.files import dysh_data
+from ..util.selection import Flag, Selection
+from . import conf
 from .sdfitsload import SDFITSLoad
 
 # from GBT IDL users guide Table 6.7
 # @todo what about the Track/OnOffOn in e.g. AGBT15B_287_33.raw.vegas  (EDGE HI data)
 # _PROCEDURES = ["Track", "OnOff", "OffOn", "OffOnSameHA", "Nod", "SubBeamNod"]
 
+pd.set_option("display.max_rows", conf.summary_max_rows)
```
mpound, line 50: Move this line to inside summary().

mpound (approved):
Nice work!

astrofle:
> This infrastructure is good to have and the initial implementation works. I can add a Conf to `plot/__init__.py` and it gets written to ~/.dysh/config/dysh.cfg. But I don't understand how the user uses it. For instance, if I edit ~/.dysh/config/dysh.cfg, say to change summary_max_rows, it gets overwritten unless I use --skip-config. However, even if I use --skip-config summary_max_rows gets reset to 1000 by fits `__init__.py`. Furthermore, if I change it on the command line with `conf.summary_max_rows=2`, it has no effect because there is no active connection to `pd.set_options`. Moving the line `pd.set_option("display.max_rows", conf.summary_max_rows)` to inside of `summary` fixes this.
> 
> It would be good to have a utility code `dysh.show_config()` that would introspect all modules and print out their conf items if they have a conf object. Not required for this PR, just an RFE.
> 
> It may take some time before we figure out how best to use this infrastructure and make it available to users.

I don't understand why the config file gets overwritten in your case. If I modify it and then run `dysh`, I get the following message:
```Python
WARNING: The configuration file already exists and seems to have been customized, so it has not been updated. Use overwrite=True if you really want to update it. [dysh.config.core]
```
and the file has not been modified.

mpound:
I just double-checked and it definitely gets overwritten with the message 
INFO: The configuration file has been successfully written to /home/mpound/.dysh/config/dysh.cfg [dysh.config.core]

src/dysh/config/core.py
```diff
@@ -0,0 +1,203 @@
...
+    seen = set()
+    for subclass in klass.__subclasses__():
+        yield subclass
+        for subsubclass in recursive_subclasses(subclass):
+            if subsubclass not in seen:
+                seen.add(subsubclass)
+                yield subsubclass
+
+
+def generate_config(pkgname="astropy", filename=None, verbose=False):
```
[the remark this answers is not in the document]
astrofle: No. Changed it.

src/dysh/fits/gbtfitsload.py
astrofle: Thanks for catching that.

astrofle:
I'll try adding a test for that and see how it goes.

astrofle:
The test works. I'll merge. If the problem persists on your end I'd be happy to look into it.

PR #537, opened 2025-03-24
no-noise-diode-getps

astrofle:
This PR:
* Adds support for observations without noise diode for `getps`.
* Adds a test for the above.

Still needs a better treatment of the system temperature when user provided.


teuben:
In the test ps.meta["TSYS"] is 1.0 (as well as in the table["TSYS"][0], but could we add a comment here why this is 1.0, or better yet, assert it's 1.0.

Documentation says it will figure out of there is no cal, and use the no_cal mode. But there is no warning. Should there not be a warning, or does this happen often enough the user "should know" this is normal.

I still believe we should also test for the mean or rms of the spectrum as I've done in vane testing. This would allow you to test the t_sys=100 (say) keyword. Viz.
```
ps = sdf4.getps(scan=23, fdnum=10, plnum=0, t_sys=100).timeaverage()
rms = ps.stats()["rms"].value  # 0.4233987230657062
assert np.isclose(rms, 0.4233987230657062, rtol=1e-7)
```
where we should have some agreement what the rtol value should be?  I was a bit sloppy and took 1e-4, I saw you took 1e-7 in the spectrum difference, which doesn't work with t_sys=100 anyways,  need a 2nd getps.

teuben:
In your README file it starts with /home/dysh/example_data/onoff-Argus/data/TGBT24B_613_15.raw.vegas which I cannot find.   I do find a TGBT24B in the example_data/mixed-fs-ps/data directory.
Of course the testdata is ok, but it seems I would be able to reproduce gbtidl workflow

(I wondered if you needed the intermediate trim, or could the two have been combined)

astrofle:
> In your README file it starts with /home/dysh/example_data/onoff-Argus/data/TGBT24B_613_15.raw.vegas which I cannot find. I do find a TGBT24B in the example_data/mixed-fs-ps/data directory. Of course the testdata is ok, but it seems I would be able to reproduce gbtidl workflow
> 
> (I wondered if you needed the intermediate trim, or could the two have been combined)

Good catch. I added the data to:
`/home/dysh/example_data/onoff-Argus/data/`

astrofle:
> In the test ps.meta["TSYS"] is 1.0 (as well as in the table["TSYS"][0], but could we add a comment here why this is 1.0, or better yet, assert it's 1.0.
> 
Added a check for TSYS value equal to one.

> Documentation says it will figure out of there is no cal, and use the no_cal mode. But there is no warning. Should there not be a warning, or does this happen often enough the user "should know" this is normal.
> 
Added an `info` level message telling the user it will not use the noise diode, and which system temperature it will use.

> I still believe we should also test for the mean or rms of the spectrum as I've done in vane testing. This would allow you to test the t_sys=100 (say) keyword. Viz.
> 
> ```
> ps = sdf4.getps(scan=23, fdnum=10, plnum=0, t_sys=100).timeaverage()
> rms = ps.stats()["rms"].value  # 0.4233987230657062
> assert np.isclose(rms, 0.4233987230657062, rtol=1e-7)
> ```
> 
Added a test for the rms and mean of the calibrated spectrum.

> where we should have some agreement what the rtol value should be? I was a bit sloppy and took 1e-4, I saw you took 1e-7 in the spectrum difference, which doesn't work with t_sys=100 anyways, need a 2nd getps.
Will start an issue so we talk about it.

Issue #511, opened 2025-03-14
init_logging with verbosity 0 (Error)


astrofle:
Initializing the logging with `verbosity=0` does not work because of this line:
https://github.com/GreenBankObservatory/dysh/blob/eea079e5ea9cf31bc4c26ed6d225627d90d2e29b/src/dysh/log.py#L136
This is `False` if `verbosity=0`.

astrofle:
Thanks to Peter for catching this.

teuben:
oops, I just had it as a comment in the code someplace. somebody actually read my code?

PR #539, opened 2025-03-25
Fix for issue 511

tchamberlin:
This resolves issue #511

Original behavior:

```
$ uv run dysh --verbosity=0 --no-banner
Traceback (most recent call last):
  File ".../.venv/bin/dysh", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File ".../src/dysh/shell/shell.py", line 99, in main
    init_logging(verbosity=args.verbosity, path=args.log, quiet=args.quiet)
  File ".../src/dysh/log.py", line 149, in init_logging
    raise ValueError("One of verbosity or level must be given!")
ValueError: One of verbosity or level must be given!
```

New behavior:

```
$ uv run dysh --verbosity=0 --no-banner

In [1]: from dysh.log import logger

In [2]: logger.error("Foo")
ERROR Foo

In [3]: logger.info("Foo")
```


astrofle (approved):
Thanks Thomas

PR #540, opened 2025-03-26
Upgrade Python linting/formatting toolchain and fix linting errors

tchamberlin:
Ruff is a linting and formatting tool for Python. It replaces black, isort, pylint, and [many other such tools](https://docs.astral.sh/ruff/faq/#how-does-ruffs-formatter-compare-to-black)

Ruff's autoformatting is generally identical to black's, but there are _some_ differences: https://docs.astral.sh/ruff/faq/#how-does-ruffs-formatter-compare-to-black

> Specifically, the formatter is intended to emit near-identical output when run over Black-formatted code. When run over extensive Black-formatted projects like Django and Zulip, > 99.9% of lines are formatted identically. When migrating an existing project from Black to Ruff, you should expect to see a few differences on the margins, but the vast majority of your code should be unchanged.

This PR migrates us to ruff and removes use of black, isort and pylint. This includes updating `.pre-commit-config.toml` to use ruff.

This PR also runs `ruff check --fix`, which auto-fixes all of the "fixable" issues that it finds. There are still a lot of linter issues (this is not specific to ruff; they were there with pylint also)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,8 +25,8 @@ repos:
       - id: check-docstring-first
       - id: check-case-conflict # Check for files with names that would conflict on a case-insensitive filesystem
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.2
+    rev: v0.11.8
     hooks:
     - id: ruff-format
-    # - id: ruff
-    #   args: ['--fix', '--exit-non-zero-on-fix']
+    - id: ruff
+      args: ['--fix', '--exit-non-zero-on-fix']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,13 +24,9 @@ repos:
           - '--fix=lf'
       - id: check-docstring-first
       - id: check-case-conflict # Check for files with names that would conflict on a case-insensitive filesystem
-  - repo: https://github.com/pycqa/isort
-    rev: 6.0.1
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.11.2
     hooks:
-      - id: isort
-        exclude: '(notebooks|attic|benchmark|testdata)/.*'
-  - repo: 'https://github.com/psf/black'
-    rev: 25.1.0
-    hooks:
-      - id: black
-        exclude: '(notebooks|attic|benchmark|testdata)/.*'
+    - id: ruff-format
+    # - id: ruff
+    #   args: ['--fix', '--exit-non-zero-on-fix']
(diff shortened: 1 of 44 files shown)


tchamberlin:
well that didn't work :)

.pre-commit-config.yaml
```diff
@@ -24,13 +24,9 @@ repos:
...
+    rev: v0.11.2
     hooks:
-      - id: isort
-        exclude: '(notebooks|attic|benchmark|testdata)/.*'
-  - repo: 'https://github.com/psf/black'
-    rev: 25.1.0
-    hooks:
-      - id: black
-        exclude: '(notebooks|attic|benchmark|testdata)/.*'
+    - id: ruff-format
```
mpound, line 30: We need to keep those excluded directories.

tchamberlin: Will do, although I'll implement this in `pyproject.toml` so the behavior is consistent between CLI and pre-commit

tchamberlin:
Some more detail on what this PR changes. Unlike most, it _should not_ be squashed and merged; the commit history needs to be retained

Each of these steps is a separate commit:
1. Migrate from `black`, `pylint`, and `isort` to `ruff` by updating `pyproject.toml` and `.pre-commit.yaml` (this also updates `uv.lock`); commit
2. Run `$ ruff format` on the whole repo; commit
3. Run `$ ruff check --fix` on the whole repo; commit
4. Create `.git-blame-ignore-revs`, which contains the hashes of the two prior commits. This prevents git blame from blaming me for a bunch of stuff that only changed due to reformatting/auto-fixing

Steps 2-4 need to be re-run every time we need to rebase onto `main`, or I make a change to `pyproject.toml` (which ideally would only be once, but I keep tweaking things). So, for posterity, this script is what I'm using to re-commit things as needed (the first commit is done by hand):

```bash
#!/bin/bash

### The "ruff format" commit
ruff format
git commit -am "Run $ ruff format"
echo '# $ ruff format' > .git-blame-ignore-revs
git rev-parse HEAD^ >> .git-blame-ignore-revs

### The "ruff check" commit
ruff check --fix
git commit -am "Run $ ruff check --fix"
echo '# $ ruff check --fix' >> .git-blame-ignore-revs
git rev-parse HEAD^ >> .git-blame-ignore-revs

### The ".git-blame-ignore-revs" commit
git add .git-blame-ignore-revs
git commit -am "Add .git-blame-ignore-revs

This will retain utility of git blame across the mass-edits in the prior commits"
```

tchamberlin:
I went ahead and got us to a clean slate in terms of linting errors. There are a few genuine bugs that I am ignoring at this point, which have been marked with TODOs. And I have disabled several rules that long term we _should_ turn back on.

But hopefully at this point we can keep the linter running in pre-commit, so no further bugs of these types will be introduced

astrofle (approved):
Looks good to me.

astrofle (changes requested):
This should bump the patch version to 0.5.1

Issue #486, opened 2025-02-10
sdf.write() failing to write


teuben:
**Describe the bug**
sdf.write() now fails to write in one of my cases

**How to Reproduce**

This used to work, but after merging in code from the main in my vane1 branch, one of the examples how fails:
```
sdf2 = GBTFITSLoad(dysh_data('AGBT21B_024_01/AGBT21B_024_01.raw.vegas'), skipflags=True)
sdf2.write("edge1/file.fits", scan=[17,18], overwrite=True) 

-> TypeError: object of type 'Column' has no len()
```

**Environment**
- Dysh version 0.4.2
- Python version 3.12
- OS Linux Kubuntu 22.04

teuben:
It works with:
```
     sdf2.write("edge1/file.fits", scan=scans, overwrite=True, flags=False) 
```
and in addition,  using skipflags=False (the default) gives the same error.

astrofle:
There are two parts to this issue. 
First, the error message uses `v` instead of `values`:
https://github.com/GreenBankObservatory/dysh/blob/ff9f5db302d765bf8f658142e2e3451476d66f74/src/dysh/fits/sdfitsload.py#L967
Second, even after fixing the above there is a problem in that `GBTFITSLoad.write()` selects only one of the arrays of flags, so when trying to add a new column `SDFITSLoad._update_binary_table_column()` finds that the number of rows being added does not match the total number of rows
https://github.com/GreenBankObservatory/dysh/blob/e63188bea16bdc9c34df91727b76123f002f8f00/src/dysh/fits/gbtfitsload.py#L2424-L2429

Since different binary tables are likely to have different number of channels (not sure if that is always the case), it is not possible to build a single array with the correct dimensions.

PR #541, opened 2025-03-26
Issue 486

astrofle:
This PR
* Fixes Issue #486 
* Adds a test for it
* Fixes another small issue when loading a file with multiple binary tables written by GBTIDL


teuben:
This seems fine, I ran it myself, but haven't merged in new upstreams.

PR #551, opened 2025-04-08
354 sdfitsload does not print the filepath

teuben:
Ensured the .filename attribute is now populated when it was not.


astrofle (approved):

teuben:
I added a note that in the test for gbtfitsload I needed a Path() to make the assert work, whereas in sdfitsload - which looks identical - this was not needed.  Some symmetry missing here, TBD.

PR #557, opened 2025-04-14
546 implement getsigref

mpound:
This implements getsigref in GBTFITSLoad/PSScan, as well as pytest thereof. Signal can be int or list, reference can be int or Spectrum.   Input tsys will override calculated tsys or tsys from reference spectrum.


teuben:
very nice, the edge data using manual getps give "identical" answers. I had to make a decision on Tsys, which is not identical to yours. I can't subtract the spectra, due to something bizarre, despite both of them returning a Spectrum. Will display and followup in test_edge.py notebook.

astrofle:
My comments thus far:
* `getsigref` says nothing if it cannot calibrate the data, and this is only clear when trying to use the `_calibrated` attribute. In this case the input data has no noise diodes.
For example:
```Python
sdf = GBTFITSLoad("/home/astro-util/HIsurvey/Session02")
pssb = sdf.getsigref(scan=296, ref=295, plnum=0, ifnum=0, fdnum=0, calibrate=True) # No error here
pssb.timeaverage() # This produces an error
```
* `getsigref` does not average in time the reference integrations before doing the calibration. This results in things like:
```Python
sdf = GBTFITSLoad("/home/astro-util/HIsurvey/Session02")
# Scan 264 has 30 integrations, scan 265 has 3 integrations.
pssb = sdf.getsigref(scan=264, ref=265, plnum=0, ifnum=0, fdnum=0, calibrate=True)
print(pssb[0]._calibrated.shape) # prints (3, 8192), so only three integrations were calibrated.

**Comment by astrofle** [comment]
for the second case, unmatched number of integrations, `getsigref` should average the integrations of the reference (if ref is an int) and use the time average to calibrate the data

**Comment by mpound** [comment]
In the first case, all integrations are blank.   This can be a problem for any calibration routine, so I implemented the solution in Scan._finish_initialization() which all Scan classes call. 

In the second case, the "decimation" of the signal scans comes from `find_nonblank_ints`.   Not sure what to do here as `find_nonblank_ints` is a necessary method call.   To do what you propose and  just use the good reference data we have to either 

a) Always average the reference data in getsigref, creating an averaged Spectrum object internally, and proceed as if the user had input a Spectrum

b) Detect if len(scanrows["ON"] != len(scanrows["OFF"])  in PSScan and short-circuit `find_nonblank_ints` (?)

**Comment by astrofle** [comment]
> In the first case, all integrations are blank. This can be a problem for any calibration routine, so I implemented the solution in Scan._finish_initialization() which all Scan classes call.
> 
> In the second case, the "decimation" of the signal scans comes from `find_nonblank_ints`. Not sure what to do here as `find_nonblank_ints` is a necessary method call. To do what you propose and just use the good reference data we have to either
> 
> a) Always average the reference data in getsigref, creating an averaged Spectrum object internally, and proceed as if the user had input a Spectrum
> 
> b) Detect if len(scanrows["ON"] != len(scanrows["OFF"]) in PSScan and short-circuit `find_nonblank_ints` (?)

Let's go with a). It makes more sense given the way the GBT observes.

For the first point. I do not understand why all the integrations are blank. The same data can be calibrated with `GBTIDL`.

**Comment by mpound** [comment]
@astrofle Latest push implements creation of refscan average and updates tests and testdata.  The example of 
```
pssb = sdf.getsigref(scan=296, ref=295, plnum=0, ifnum=0, fdnum=0) 
```
raises the newly implemented Exception 'no data left to calibrate' in `_finish_initialization()` because an integration is completely blanked.  Arguably this is too strict a test, so we need to figure out what to do in this case because at that point the PSScan is being constructed and we don't want a bad PSScan getting into the ScanBlock

astrofle:
> ```python
> from dysh.fits.gbtfitsload import GBTFITSLoad
> sdf = GBTFITSLoad("/home/astro-util/HIsurvey/Session02")
> # Scan 264 has 30 integrations, scan 265 has 3 integrations.
> pssb = sdf.getsigref(scan=264, ref=265, plnum=0, ifnum=0, fdnum=0, calibrate=True)
> print(pssb[0]._calibrated.shape)
> ```

This is happening because there are no "ON" rows (no noise diode in this example). Let's get this into main and fix that later.

astrofle (approved):
Thanks a lot Marc!
(great tests!)

astrofle:
this PR should bump the minor version

PR #566, opened 2025-04-21
Iplotter selectregions

etsmit:
Can now select regions with `spec.plot(select=True)` which is the default. Add a region with "a" then click and drag on the spectrum. Once you are satisfied with your selections, press "e" to mark the spectrum and print out a `include=...` rule for the baseline call.


etsmit:
The failures are not happening locally for me. The tests here must be returning ps.spectral_axis as a numpy array somehow.

astrofle:
> The failures are not happening locally for me. The tests here must be returning ps.spectral_axis as a numpy array somehow.

It is broken for me. It is not doing the range selection appropriately:

Screenshot 2025-04-21 at 9 51 19 AM

astrofle:
It works if you have an interactive backend. I think that could be the cause of the problem.

etsmit:
Are you able to use it correctly then, with an interactive backend? 

The pytests fail because it is default plotting in channel space, which makes the spectral axis just a np.float64 range instead of an actual SpectralAxis object with units.

PR #568, opened 2025-04-21
Another option for interactive region selection

astrofle:
Another take at interactive selection.

```Python
%matplotlib ipympl # if needed
s = Spectrum.fake_spectrum()
s.plot()
# Draw regions by clicking an moving things around.
# Once you are done moving and adding regions, the selection is available in the 
# s.get_selected_regions() property.
# So, the following should work (though I have not checked it does what it should):
s.baseline(1, exclude=s.get_selected_regions())
```

### DIFF
--- a/src/dysh/plot/specplot.py
+++ b/src/dysh/plot/specplot.py
@@ -85,6 +85,7 @@ class SpectrumPlot:
         self._figure = None
         self._axis = None
         self._title = self._plot_kwargs["title"]
+        self._selector: InteractiveSpanSelector = None
 
     # def __call__ (see pyspeckit)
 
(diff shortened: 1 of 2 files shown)


etsmit:
This is all very nice, thanks! You could use "include" as well. The user just has to be cognizant of whether or not they are selecting regions of the spectrum they do or don't like.

I would like to see the button be more explicit and say "Clear Selected Regions", but that is a mouthful. I just worry when there are more features and a user sees the "Clear" button and not sure what it actually clears. Not sure what the solution is. 

Other issues which don't necessarily block merging, but worth noting:
- actually running the baseline command doesn't update the plot (did it do this before? even with `remove=True`?)
- running the plot() command again on the same spectrum object loses the reference to the first one. So, even if you close out the second plot, then add selections to the first plot, you can't get anything from get_selected_regions.

etsmit (approved):

PR #569, opened 2025-04-22
Q8 test updates

astrofle:
This PR:
* Fixes (?) a potential issue with one of the tests for the `Observatory` class (`astropy` recommends refreshing the cache if an observatory could not be found -- which we saw at least once recently).
* Checks for `UserWarning` in tests that involve modifying SDFITS columns. Mainly to clean up the output of the tests.


mpound (approved):
Thanks for this.

PR #576, opened 2025-04-30
feat: getnod now accepts a system temperature for each beam and/or scan

astrofle:
This PR:
* Adds `gbtfitsload._parse_tsys` function, which takes as input a system temperature value and a list of scan numbers and returns a dictionary, where the keys are the scan numbers and the values the system temperature for each beam (if applicable).
* Uses the above function in `GBTFITSLoad.getnod`, so the users can provide a system temperature for each beam and/or scan number.
* Implements tests for the above.
* getnod can use different pairs of fdnums for every scan pair.


teuben (commented):
a["FEEDXOFF"] == 0.0   should ideally   be   using np.isclose()

In the new code this is now

         mask = mask & (self["FEEDXOFF"] == 0.0) & (self["FEEDEOFF"] == 0.0)

teuben:
Description of the `t_sys` keyword in getnod doesn't match what I see in the code when callled it like:
```
     nodsb1 = sdf.getnod(scan=62, ifnum=0, plnum=0, t_sys=[[1], [2]])
```
and are these not TSYS values? Better to use something like 50 or 100, but not things that look like 0 or 1 based integers.

teuben:
thanks for merging the nod_beams codes. how did that ever happen.

teuben:
taggeed   - typo

teuben (approved):
ok now.

Issue #578, opened 2025-05-01
gettp grabs the wrong data


astrofle:
**Describe the bug**
Using `GBTFITSLoad.gettp` grabs the wrong data in some cases.

**How to Reproduce**
```Python
from dysh.fits.gbtfitsload import GBTFITSLoad

sdf = GBTFITSLoad("/home/astro-util/HIsurvey/Session02")
tp = sdf.gettp(scan=263,ifnum=0,plnum=0,fdnum=0)
for i in range(len(tp[0])):
    print(tp[0].meta[i]['SCAN'],tp[0].meta[i]['OBJECT'])
```
This will print scan 220 and 3C286, when the scan selected was 263 and the source should have been U8091.

Thanks to Marc for reporting this.

**Environment**
- Dysh version 0.4.3
- Python version
- OS

astrofle:
The initial selection grabs the correct data:
```Python
(scans, _sf) = sdf._common_selection(fdnum=0, ifnum=0, plnum=0, apply_flags=True, scan=263)
```
Shows data for scan 263.

astrofle:
Next selection still gets the correct data:
```Python
from dysh.util import select_from
_df = select_from("FITSINDEX", 0, _sf)
```
Shows the same `pandas.DataFrame` as above.

PR #579, opened 2025-05-01
Issue 578

astrofle:
This PR:
* Fixes issue #578 
* Changes the `__len__` of `ScanBase` to use the number of integrations
* Changes the `scans` keyword in `GBTFITSLoad.summary` to `scan` for consistency
* Bumps the package version to 0.4.5


src/dysh/fits/gbtfitsload.py
```diff
@@ -1059,30 +1059,35 @@ def gettp(
...
                 # sig and cal are treated specially since
                 # they are not in kwargs and in SDFITS header
                 # they are not booleans but chars
                 if sig is not None:
                     _sifdf = select_from("SIG", TF[sig], _sifdf)
+                if bintable is None:
+                    bintable = set(_sifdf["BINTABLE"])
+                    # I do not know if this is possible, but just in case.
+                    if len(bintable) > 1:
+                        raise TypeError("Selection crosses binary tables.")
```
mpound, line 1082: In the long run, we  may wish to find a way to allow selections that cross binary tables, if it turns out to be a significant use-case.

mpound (approved):
Thank you for quickly finding the solution.  I love waking up to find bugs fixed!

PR #582, opened 2025-05-02
reset patch number after minor version bump

astrofle:
The path number should be reset after bumping the minor version. Similarly, when bumping the major version, the minor and patch should be reset. Source: https://semver.org/#semantic-versioning-specification-semver


mpound (approved):
Ah, that makes more sense.

PR #586, opened 2025-05-14
Iplotter bshape

etsmit:
If remove=True in spectrum.baseline(), removes the fitted baseline and updates the plot. If remove=False, computes the baseline fit and plots it over the spectrum.


astrofle:
When working in jupyterlab, using `Spectrum.baseline` with `remove=False` and then plotting does not show the baseline. For example:
```Python
s = Spectrum.fake_spectrum()
s += 10
s.baseline(1, model="poly", remove=False)
s.plot()
```
Notice that I am using the non-interactive backend in the example.
However, this does show the baseline:
```Python
s = Spectrum.fake_spectrum()
s += 10
s.plot()
s.baseline(1, model="poly", remove=False)
```
This seems like a specific issue with how jupyter executes cells.

astrofle:
Saving the figures works well in these cases:
* `s.plot(); s.savefig("test.png");`
* `s.plot(xaxis_unit="km/s"); s.savefig("test_kms.png");`
* `s.plot(show_header=False); s.savefig("test_nohead.png");`

astrofle:
If working on a jupyter notebook, calling `Spectrum.baseline` will update the plots, but the update will take place in the last instance of the figure, so it could be updating a plot X cells above. This should be documented.
Personally, I'd prefer to have the plot update in the cell where the update takes place, but this is just my opinion.

astrofle:
I think this is ready for merging as is. We should definitely work on documenting this before adding more features.

astrofle (approved):
Thanks a lot Evan! Great work and so fast.

astrofle:
This should update the package version to 0.6.0. This is set in `src/dysh/__init__.py`

PR #593, opened 2025-05-22
docs: update read the docs workflow to use uv

astrofle:
After we removed the `requirements.txt` file our read the docs builds have been failing. This PR updates the read the docs workflow to use `uv` based on this: https://github.com/astral-sh/uv/issues/10074

I tested the results on a separate read the docs instance, and the changes fix the issue:
https://dysh-test.readthedocs.io/en/rtd-uv/tutorials/examples/hi_survey.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,13 +7,17 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
-    os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


tchamberlin (approved):
Looks good to me. Hopefully readthedocs adds first-party support for uv in future, but this seems fine for now

PR #601, opened 2025-05-27
W band example

astrofle:
This PR:
* Adds an example notebook for W-Band observations
* Fixes an issue where astropy would log info level messages
* Fixes two issues with `Spectrum.average`: it wasn't masking right, and it wasn't keeping the history
* Style update: when plotting, the header won't show the source name if there is a user provided figure title
* Minor clean up of the Nodding notebook
* Fixes an issue with the inversion of include regions during baseline fitting


mpound (approved):
Nice job!
I made minor changes to a few files.   One comment to resolve on a parameter description.  Thanks,

PR #605, opened 2025-06-11
561 apply baseline solutions to the integrations in a scanscanblock

mpound:


astrofle (approved):
Thanks a lot Marc! I enjoyed playing with it.
I made some small changes to the docstrings and comments, fixed the issue with the ascending order of the domain and spectral_axis during the model check (it would fail, because the domain will be descending if the spectral_axis is descending), and added tests with the data values just to make sure that does not break.
I also fixed a bug related to the plotter I had not caught before.

src/dysh/spectra/scan.py
```diff
@@ -129,10 +129,10 @@ def __init__(
         self._apply_flags = apply_flags
         self._observer_location = observer_location
         self._bunit_to_unit = {"ta": u.K, "ta*": u.K, "jy": u.Jy, "counts": u.ct}
-        # possible @todo: create a BaselineableMixin that Spectrum, ScanBlock, and ScanBase inherif from.
+        # possible @todo: create a BaselineableMixin that Spectrum, ScanBlock, and ScanBase inherit from.
         self._baseline_model = None
         self._subtracted = (
-            False  # This is False iff baseline_model is None so we technically don't need a separate boolean.
+            False  # This is False if baseline_model is None so we technically don't need a separate boolean.
```
mpound, line 135: I did mean "iff" here  "if and only if"

PR #606, opened 2025-06-11
564 performance testing 1

teuben:
This adds a framework to test cpu and memory, time tag your code, so I/O overhead can be estimated as well (but we aware of I/O caching).

Current scirpts and Q8 baseline results are in $DYSH/benchmark and the q8stats directory in there.


mpound (approved):
I approve the final outputs, but wait until Pedro's review to merge.

src/dysh/util/timers.py
```diff
@@ -0,0 +1,279 @@
...
+    data_units : list
+         List of units names of the extra columns.
+
+    data_types : list
+         List of data types of the extra columns.
+
+    args: dict
+         This dictionary controls a number of common variables used in dysh benchmarking.
+
+         out        : output filename (astropy Table). Default it none is written.
```
etsmit, line 45: "Default it none is written"?

etsmit (approved):
Looks good. My only feedback is that a small one-liner can be added to the top of each `benchmark/bench*.py` file explaining what aspect of dysh this file focuses on. Or, a blurb in the README quickly explaining each file. Some are obvious, like getps and gbtfitsload, and some are less obvious, like edge and skel.

Issue #614, opened 2025-06-18
conf directory and contents not created when installing dysh


astrofle:
**Describe the bug**
While going through the calseq tutorial Emily M. reported an issue while trying to correct for atmospheric opacity.
The issue is that the `GBTGainCorrection` class cannot find the file `gain_correction_table = get_project_configuration() / "gaincorrection.tab"`. Looking at the filesystem, the file indeed does not exist.

**How to Reproduce**

For this to error, you must be working from a fresh dysh install, for example:
```bash
conda create -n dysh-test python==3.10
conda activate dysh-test
pip install "dysh[nb] @ git+https://github.com/GreenBankObservatory/dysh"
```
Then, run the following lines from `dysh`:
```Python
from astropy.table import QTable
from dysh.util import get_project_configuration
gain_correction_table = get_project_configuration() / "gaincorrection.tab"
QTable.read(gain_correction_table, format="ascii.ecsv")
```

**Environment**
- Dysh version 0.7.0
- Python version 3.10
- OS

Issue #615, opened 2025-06-18
dysh.util.get_project_root is innacurate when dysh is installed


astrofle:
**Describe the bug**
When `dysh` is installed, only the files in `src/dysh` get put into the `site-packages` directory of the Python lib. This changes the directory structure, relative to what we normally see if working from the repository. So that going up three parents, as `dysh.util.get_project_root` does, lands nowhere near where we would like it to be.

**How to Reproduce**
Install ``dysh``, like a user:
```bash
conda create -n dysh-test python==3.10
conda activate dysh-test
pip install "dysh[nb] @ git+https://github.com/GreenBankObservatory/dysh"
```
Launch ``dysh`` and execute:
```Python
from dysh.util import get_project_root
get_project_root()
```
will say something like:
`PosixPath('/Users/psalas/miniconda3/envs/dysh-w/lib/python3.10')`
but the code is under:
`PosixPath('/Users/psalas/miniconda3/envs/dysh-w/lib/python3.10/site-packages/dysh')`
which is correctly reported by `importlib.resources.files("dysh")`.

**Environment**
- Dysh version 0.7.0
- Python version 3.10
- OS

PR #620, opened 2025-06-20
issue 614

astrofle:
This PR fixes issues #614 and #615 by
* Renaming conf/ to src/dysh/data/
* Updating `get_project_root` to use `importlib` and point to src/dysh/
* Updating the tests to directly use `get_project_testdata` rather than starting from `get_project_root`

In the end I decided not to move testdata/ into src/dysh/.


mpound (approved):
Good! Thanks

PR #623, opened 2025-06-24
617 scanblocksdwrite seems to write the first scan twice

teuben:
The first scan was written twice for a  scanblock.  Copy and paste can create identical messages, these were "fixed up" as well.


src/dysh/spectra/scan.py
```diff
@@ -944,6 +945,7 @@ def write(self, fileobj, output_verify="exception", overwrite=False, checksum=Fa
         defaultkeys = set(s0._meta[0].keys())
         datashape = np.shape(s0._calibrated)
         tablelist = [s0._meta_as_table()]
```
mpound, line 947: Wouldn't replacing this line with `tablelist = []` accomplish the same thing?

mpound: The nrows method is preferable because it can be used in the logger.debug.

PR #627, opened 2025-06-26
626 return plot obj

etsmit:
Added an __init__ call to the beginning of plot(), as calling this consecutive times (e.g. in practically every example jupyter notebook) would not update any kwargs (ymin, xaxis_unit, etc.). Not sure if this is good coding practice.

### DIFF
--- a/src/dysh/plot/specplot.py
+++ b/src/dysh/plot/specplot.py
@@ -85,6 +85,7 @@ class SpectrumPlot:
     def __init__(self, spectrum, **kwargs):
         self.reset()
         self._spectrum = spectrum
+        self._sa = spectrum._spectral_axis
         self._set_xaxis_info()
         self._plot_kwargs.update(kwargs)
         self._plt = plt
@@ -154,7 +155,6 @@ class SpectrumPlot:
 
         s = self._spectrum
 
-        self._sa = s.spectral_axis
         lw = this_plot_kwargs["linewidth"]
         xunit = this_plot_kwargs["xaxis_unit"]
         yunit = this_plot_kwargs["yaxis_unit"]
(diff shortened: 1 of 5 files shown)


src/dysh/plot/specplot.py
```diff
@@ -135,6 +136,7 @@ def plot(self, show_header=True, select=True, interactive=True, **kwargs):
         **kwargs : various
             keyword=value arguments (need to describe these in a central place)
         """
+        self.__init__(self._spectrum, **kwargs)
         if interactive:
             plt.ion()
```
mpound, line 141: perhaps this should be `self._plt` for consistency.

src/dysh/plot/specplot.py
```diff
@@ -135,6 +136,7 @@ def plot(self, show_header=True, select=True, interactive=True, **kwargs):
...
         plt.rcParams["font.family"] = "monospace"
```
mpound, line 142: same here.

src/dysh/plot/specplot.py
```diff
@@ -442,6 +448,22 @@ def get_selected_regions(self):
         regions = self._selector.get_selected_regions()
         return [tuple(np.sort([np.argmin(abs(p - self._sa.value)) for p in r])) for r in regions]
 
+    def freex(self):
+        self._freezex = False
+        # This line (and the other in specplot.py) will have to be addressed when we
+        # implement multiple IF windows in the same plot
+        self._axis.set_xlim(np.min(self._sa).value, np.max(self._sa).value)
```
mpound, line 455: SpectralAxis have a min()/max() methods, so these could be self.sa.min().value and self.sa.max().value

src/dysh/plot/specplot.py
etsmit: I'm not sure I understand? Are you referring to lines 141 and 142?

etsmit: Ahh, I got it.

mpound:
@etsmit Just noticed that there are tests where spectrum.plot() is still called with show=False instead of interactive=False, e.g. test_spectrum.py.   Can you ensure these are all taken care of?  They open up mucho windows in pytest.

etsmit:
Thanks for catching that. I fixed it in a few other places but didn't realize how many more there were.

Issue #632, opened 2025-07-02
iplotter baseline model does not show when xaxis_unit='chan'


etsmit:
Self-explanatory.

The baseline model is probably showing up at the frequency values of the spectral axis, and not at the channel values.

Issue #630, opened 2025-07-02
help for Spectrum.baseline() does not match the API


teuben:
**Describe the bug**

API says: sp.baseline(degree, exclude=None, include=None, color='k', **kwargs)

the docs describe more (that's nice to cover some of the **kwargs), but at least color= should be described in the docs.

**How to Reproduce**

sp.baseline?

**Environment**
- Dysh version 0.7
- Python version
- OS

teuben:
Should we not say something that if a sp.plot() is up, it will overplot?  Only the color= gives a hint this is happening.  If color='' would it skip over-plotting?

teuben:
color= says   `The color to plot the baseline model, if not removing from the spectrum. Can be any type accepted by matplotlib.`

the way I read this, it implies if the baseline is subtracted, no plot is done. Correct ?

etsmit:
Yes, if remove=True then it will not plot the baseline model. It will subtract the model out and replot the power spectrum. I can add a sentence to the description.

teuben:
I think the comma threw me off. Two sentences may make it more clear.

etsmit:
```
color : str
    The color to plot the baseline model, if remove=False. Can be any type accepted by matplotlib.
model : str
    One of 'polynomial', 'chebyshev', 'legendre', or 'hermite'
    Default: 'chebyshev'
fitter : `~astropy.modeling.fitting.Fitter`
    The fitter to use. Default: `~astropy.modeling.fitting.LinearLSQFitter` (with `calc_uncertaintes=True`).
    Be careful when choosing a different fitter to be sure it is optimized for this problem.
remove : bool
    If True, the baseline is removed from the spectrum.
    If False, the baseline will be computed and overlaid on the power spectrum. Default: False
```

teuben:
ok,but I would leave out the word "power' before spectrum. It's overloaded (power spectrum makes me think FFTs). I think spectrum is sufficient.  Do we use "power spectrum" in other places?

PR #633, opened 2025-07-07
632 baseline model show

etsmit:
Closes [issue 632](https://github.com/GreenBankObservatory/dysh/issues/632) and [issue 630](https://github.com/GreenBankObservatory/dysh/issues/630).


teuben (approved):
thanks Evan, this "just works". Checked freq, km/s and chan. Will come in handy with the staff_training

PR #647, opened 2025-07-11
OTF docs

astrofle:
Peter and I forgot to add the on the fly example notebook to the documentation, and bump the patch version. This PR does that.


mpound:
- In the HTML version of the notebook after the downloading cell there is this 
```
Failing to retrieve example TGBT17A_506_11.raw.vegas.A.fits 
None
```
- a typo after the equation for `T_A`, should read "T_sys is..."   (missing the s)
- The example of Scanblock.subtract_baseline doesn't use a ScanBlock, but rather a Scan.  Should add the ScanBlock version even if it shows that its not always the best choice?
```
 sb0[len(sb0)//2].subtract_baseline(spec_avg.baseline_model)
```
- the loop over integrations baseline example "Answer" could be sped up by creating a Spectrum once and replacing its data with `_c` in the loop.   Peter has a script version which does this. 
- gbtgridder section, change to: "be sure to be in touch with the gbtgridder developers"
- plotting section, you say "to keep equal aspect ratios for the axes (not really needed since the data is already a square)."  but the cell has aspect='auto' rather than aspect='equal'   

Good job!

mpound (approved):
A few minor doc changes.

src/dysh/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Top-level package for dysh."""
 
-__version__ = "0.7.0"
+__version__ = "0.7.1"
```
mpound, line 3: This will need to be 0.7.2 as my merge claimed 0.7.1

teuben: probably 0.7.2  by the time.....

Issue #649, opened 2025-07-14
summary won't show if it is not the last line in a cell


astrofle:
**Describe the bug**
When using `GBTFITSLoad.summary()`, it won't show up if one has a code line after calling it in a code cell in a notebook.

**How to Reproduce**
In a notebook do this in a single cell, and no summary will be displayed:
```Python
sdf = GBTFITSLoad(...)
sdf.summary()
sdf.getps()
```

**Environment**
- Dysh version 0.7.0
- Python version
- OS

teuben:
I think it's a  notebook feature  :-)

sdf.summary() returns a dateframe, and since it's the last thing in the cell, returns it. 

If the cell was:
```
sdf = GBTFITSLoad(...)
sdf.summary();
```
it would also not display the summary.

PR #656, opened 2025-07-21
fix: issue #649

astrofle:
Split summary into two functions summary and get_summary. get_summary does the heavy lifting of creating the DataFrame. summary now calls get_summary and displays the result using dysh.util.core.show_DataFrame. This will show the summary even if the call is not the last one in a cell.
It also adds an argument for the maximum number of rows and hides the index by default.


mpound (changes requested):
See what you can find out about the the max_rows behavior the same in different platforms.

astrofle:
> See what you can find out about the the max_rows behavior the same in different platforms.

It was the use of a Styler object that was making the difference. Calling `DataFrame.to_html` or `DataFrame.to_string` directly produces the same result. Now `GBTFITSLoad.summary()` shows the same type of ellipsis on the three platforms.

Issue #648, opened 2025-07-14
sdf.summary() scan ordering does not always match GBTIDL


dbautista98:
**Describe the bug**

Using the GBO installation of dysh to display the summary of some data does not order the scans in the same order that GBTIDL does. I expected the scans to be ordered by increasing scan number, but for some observing sessions, the scans are ordered differently. 

**How to Reproduce**

Session `TGBT25A_607_11` is the only session that I've experienced this behavior for, but there may be others. 

Dysh: 
```
sdf = GBTOffline("TGBT25A_607_11")
sdf.summary()

    SCAN OBJECT VELOCITY   PROC  PROCSEQN RESTFREQ DOPFREQ # IF # POL # INT # FEED     AZIMUTH   ELEVATIO
0     33  3C286      0.0  OnOff         1     2.35   1.225    3     2    93      1   234.93022  77.129126
1     34  3C286      0.0  OnOff         2     2.35   1.225    3     2    93      1  238.048463  76.578554
2     35  3C286      0.0  OnOff         1     2.35   1.225    3     2    93      1  239.297507  75.930807
3     36  3C286      0.0  OnOff         2     2.35   1.225    3     2    93      1  242.114389  75.354417
4      6  3C196      0.0  OnOff         1     2.35   1.225    3     2    93      1  304.876488  40.689247
5      7  3C196      0.0  OnOff         2     2.35   1.225    3     2    93      1  306.079487  40.145232
6      8  3C196      0.0  OnOff         1     2.35   1.225    3     2    93      1  305.318338  39.522565
7      9  3C196      0.0  OnOff         2     2.35   1.225    3     2    93      1  306.531063  38.981578
8     20  3C295      0.0  OnOff         1     2.35   1.225    3     2    93      1   21.240329  75.055131
9     21  3C295      0.0  OnOff         2     2.35   1.225    3     2    93      1   20.415099  75.285874
10    22  3C295      0.0  OnOff         1     2.35   1.225    3     2    93      1   17.240679  75.525653
11    23  3C295      0.0  OnOff         2     2.35   1.225    3     2    93      1   16.271273  75.712528
```

Comparing to GBTIDL:
```
offline, "TGBT25A_607_11"
summary

  Scan           Source      Vel    Proc Seq    RestF nIF nInt nFd     Az    El
-------------------------------------------------------------------------------
     6            3C196      0.0   OnOff   1    1.225   3   93   1  304.8  40.9
[...]

Issue #638, opened 2025-07-08
.summary() not working well for multiple fits files


swapnaneeldey:
When running the .summary() command for a "dysh.fits.gbtfitsload.GBTFITSLoad" type file, which has multiple fits files loaded (2 in this case), the table shows incorrect number of sources, and possibly integration time.
I would expect the table for each of the fits file to be simply combined, but .summary() seems to do the following 2 things:
1) Finds common sources in both fits and adds their integration time.
2) Limits the final table to include only up to 35 rows, when it should include 70 in this case (since both fits files contain 35 scans).

**How to Reproduce**
sdfits = GBTFITSLoad("/path/to/directory/containing/multiple/fits")
sdfits._sdf #should show multiple fit file names in a list
sdfits.summary()

**Another doubt**

I also wonder if this is just an issue in the .summary(), since I do the following with the same sdfits
ps0_obj = sdfits.getps(object = "BC17", plnum=0, ifnum=0, fdnum=0).timeaverage()
ps1_obj = sdfits.getps(object = "BC17", plnum=1, ifnum=0, fdnum=0).timeaverage()
ps_obj = ps0_obj.average(ps1_obj)

ps_obj.set_frame("icrs")
new = ps_obj.smooth("gaussian", 33.3)
new.set_frame("icrs")
new.plot(xaxis_unit = "km/s",yaxis_unit="K", ymin=0, ymax=.2, xmin = -800, xmax = 2500) 

However, it presents a slightly different plot than the one I recreate in GBTIDL.

<img width="966" height="531" alt="Image" src="https://github.com/user-attachments/assets/04e8f0c2-5b0e-42e6-b953-93ff649ac716" />
<img width="2622" height="1154" alt="Image" src="https://github.com/user-attachments/assets/939150f4-4dfe-4b54-9f9f-31240495071d" />

astrofle:
Dear @swapnaneeldey ,
Thanks for reporting this issue.
Could you share the version of dysh you used?
Thanks,

swapnaneeldey:
Hello @astrofle,
I am using 0.7.0

astrofle:
Could you point me to the files you are trying to load. Thanks.

swapnaneeldey:
/users/sdey/blue_blobs/24B_fits/BC17/

This is the path on the GBT machine. Please let me know if you are not able to access this.

astrofle:
and the GBTIDL code?

swapnaneeldey:
[...]

PR #659, opened 2025-07-22
Summary issues

astrofle:
Fixes issues #648 and #638, and adds functionality for users to provide the columns they want in the summary.
Not all the columns are supported at this point.

### DIFF
--- a/src/dysh/plot/specplot.py
+++ b/src/dysh/plot/specplot.py
@@ -157,19 +157,19 @@ class SpectrumPlot:
         s = self._spectrum
 
         lw = this_plot_kwargs["linewidth"]
-        xunit = this_plot_kwargs["xaxis_unit"]
+        self._xunit = this_plot_kwargs["xaxis_unit"]  # need to kick back a ref to xunit for baseline overlays
         yunit = this_plot_kwargs["yaxis_unit"]
-        if xunit is None:
-            xunit = str(sa.unit)  # noqa: F821
+        if self._xunit is None:
+            self._xunit = str(sa.unit)  # noqa: F821
         if "vel_frame" not in this_plot_kwargs:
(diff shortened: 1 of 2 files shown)


src/dysh/fits/gbtfitsload.py
```diff
@@ -437,128 +437,199 @@ def get_summary(self, scan=None, verbose=False):
...
-        ]
-
-        # In the process, some columns get cast to floats or others. Make sure we cast them
-        # back to an appropriate data type before return.
-        col_dtypes = {"SCAN": int, "PROCSEQN": int}
-        uncompressed_df = None
+
+        from collections import namedtuple
+
+        col_def = namedtuple(
```
mpound, line 465: This is very clever and forward-looking.

src/dysh/fits/gbtfitsload.py
```diff
@@ -437,128 +437,199 @@ def get_summary(self, scan=None, verbose=False):
...
+            col_defs_set = set(col_defs.keys())
+            diff = show_set - col_defs_set
+            if len(diff) > 0:
+                raise ValueError(f"Column(s) {diff} are not handled yet.")
+            # No duplicate columns.
+            if len(show_set) < len(show):
+                logger.warning("show contains duplicated column(s). Removing them. Column order won't be preserved.")
+                show = list(show_set)
+            # No single SCAN column.
+            if show == ["SCAN"]:
```
mpound, line 578: Why is SCAN being treated specially here?    Any other column is allowed to fly solo.

astrofle: I can't figure out how to allow for SCAN and/or PROJID as only columns and not have to add some very specific lines of code in there.

PR #660, opened 2025-07-23
dysh_data_url

astrofle:
Add DYSH_DATA_URL as a system variable to be used by `dysh_data`. It overrides the default url.


teuben (commented):
I had just commited dysh_data for the shell in my 641. basicallyu a conflict with no conflict, but to be resolved.

the DYSH_DATA_URL is a good idea, it does make some assumptions, e.g. at GBO /home/sdfits is used, and if the URL is different (e.g. at UMD), will it still work, or would you expect it to work?

PR #673, opened 2025-07-31
646 improve performance of loading data with and without flags ps review

astrofle:
Here are my changes. I also added a tutorial, which I felt was sorely missing given all the nice features you've added.


astrofle:
Sorry about the mess with the commit history. I wanted to check some code from main, and it messed it up.

Issue #653, opened 2025-07-17
wrong reduction in SDFITSLoad._column_mask


astrofle:
The code in: https://github.com/GreenBankObservatory/dysh/blob/febb1d74ab0bdc4ce30def095b4ccb8cb90fb13f/src/dysh/fits/sdfitsload.py#L656
will produce a wrong combination of masks as is. The correct version would be:
```Python
_mask = np.logical_and.reduce(_mask)
```
Making a note here so I don't forget.

PR #679, opened 2025-08-05
Fix issue 653

astrofle:
fix: wrong reduction operation (issue #653 )


mpound:
Good catch. The original code would only work if the mask array was [2,n] or [3,n].    Seems our tests never flexed this.  Should we add a regression test?

astrofle:
> Good catch. The original code would only work if the mask array was [2,n] or [3,n]. Seems our tests never flexed this. Should we add a regression test?

added

mpound (approved):

Issue #654, opened 2025-07-19
Units in baseline exclude= not always working


teuben:
**Describe the bug**

Units in the exclude=   (and perhaps also include=) are not handled properly, possibly related to the slice() operator

**How to Reproduce**

The standard dysh_data(example="test1") has been used for this, since we are also using it in the staff training.

```
from astropy import units as u
from dysh.fits import GBTFITSLoad
from dysh.util.files import dysh_data


f1 = dysh_data(example="test1")
sdf1 = GBTFITSLoad(f1)
sdf1.summary()

sp1 = sdf1.getps(fdnum=0, ifnum=0, plnum=0).timeaverage()
sp2 = sp1[slice(1.42*GHz, 1.38*GHz)]
sp2.plot()
sp3 = sp2.smooth("box", 32)
sp3.plot(xaxis_unit="km/s")


sp4 = sp3[slice(1500*kms,7500*kms)]
sp4.plot(xaxis_unit="km/s")

#e = [(4000*kms,5000*kms)]       # this will give an obvious bias -0.01017421
#e = [(5000*kms,4000*kms)]       # this is ok   0.00052977
e = [(1.397*GHz, 1.403*GHz)]     # this is ok   0.0002324
sp4.plot()
sp4.baseline(1,exclude=e, model="polynomial", remove=False)  # watch the fit looks ok
sp4.baseline(1,exclude=e, model="polynomial", remove=True)   # watch the spectrum looks baselined ok
sp4.plot(xaxis_unit="chan")
sp4[:200].stats()

```

perhaps related is how slices are handled when the axis is not ordered in the same way as the min and max in the slice. Or is that part a pilot error

```
s1 = slice(4000*kms,5000*kms) 
s2 = slice(5000*kms,4000*kms)
s3 = slice(1.397*GHz, 1.403*GHz)

print(len(sp4.flux))        # 565
print(len(sp4[s1].flux))    # 94  - bad in baseline
print(len(sp4[s2].flux))    # 0   - but good in baseline
print(len(sp4[s3].flux))    # 0   - but good in baseline

```
so this is as if the exclude is not using slices, but still winds up making the same logic error.

Essentially in the bad case, like in wrong slicing, it doesn't actually exclude the line, it uses the whole spectrum, and thus gives that bias. Too subtle to see in the original notebooks since the whole spectrum and unsmoothed was used.

**Environment**
- Dysh version 0.7.0
- Python version  3.11.x
- OS kubuntu 24

PR #680, opened 2025-08-05
Fix issue 654

astrofle:
Fix for issue #654 

It makes sure the lower and upper bounds of the exclude and include regions match what `specutils` expects.


teuben (commented):
in test_spectrum a few absolute are used, e.g. 574 and 853, but also 1402.3 and 1402.5. I would prefer to see them only being defined once, and then reused as a variable later on.  For the channel numbers ch_low and ch_upp are already available.

teuben (approved):

PR #683, opened 2025-08-06
getfs without noise diode

astrofle:
This PR adds support for `getfs` when there are no noise diodes firing. It also adds support for `smoothref`.

It does not solve the Argus calibration issues.

### DIFF
--- a/src/dysh/spectra/tests/test_scan.py
+++ b/src/dysh/spectra/tests/test_scan.py
@@ -526,6 +526,32 @@ class TestFSScan:
         assert fs.stats()["mean"].value == pytest.approx(0.19299398411039015)
         assert fs.stats()["rms"].value == pytest.approx(5.4871938402480795)
 
+        # nocal.
+        fs_sb = sdf.getfs(scan=20, ifnum=0, plnum=0, fdnum=0, fold=True, smoothref=256, nocal=True)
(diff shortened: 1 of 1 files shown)


src/dysh/fits/gbtfitsload.py
```diff
@@ -1866,11 +1867,13 @@ def getfs(
             If 'ta*' or 'jy' the zenith opacity must also be given. Default:'ta'
         zenith_opacity: float, optional
                 The zenith opacity to use in calculating the scale factors for the integrations.  Default:None
-        observer_location : `~astropy.coordinates.EarthLocation`
-            Location of the observatory. See `~dysh.coordinates.Observatory`.
-            This will be transformed to `~astropy.coordinates.ITRS` using the time of
-            observation DATE-OBS or MJD-OBS in
-            the SDFITS header.  The default is the location of the GBT.
+        t_sys : float, optional
```
mpound, line 1870: float or ~np.ndarray.  Then explain the axes of the ndarray.   (At some point we need to make the type hints and docstrings consistent for all calibration routines.   I'll file an issue unless you want to do it in this PR).

mpound (changes requested):
Good job! Should add test for input tsys and doc for tsys as a numpy array.

mpound (approved):

PR #686, opened 2025-08-11
flux calibration routines

astrofle:
This PR adds:
* `Calibrator` and `CalibratorTable` classes to hold calibrator information
Will add:
* `gettcal` function to compute the calibrator temperature given a "valid" scan number


src/dysh/data/calibrators.json
```diff
@@ -0,0 +1,70 @@
+{
+    "Objects" :
```
mpound, line 2: Some of the names contain unicode chars which makes `from_name()` will fail.   In ipython if you type 
```
c=Calibrator()
c.calibrator_table.data
```
you will see the unicode.  I see 0133-3629, 0918-1205, 1720-0058, 2214-1701are unicode.

src/dysh/util/calibrator.py
```diff
@@ -0,0 +1,283 @@
...
+        """
+
+        if calibrator_table is None:
+            calibrator_table = CalibratorTable()
+
+        if scale not in calibrator_table.valid_scales():
+            raise ValueError(f"Unrecognized scale. Valid scales: {', '.join(calibrator_table.valid_scales())}")
+        if name not in calibrator_table.valid_names(scale):
+            raise ValueError(
+                f"Unrecognized calibrator name. Valid names: {', '.join(calibrator_table.valid_names(scale))}"
```
mpound, line 207: insert {name} after 'Unrecognized calibrator name '

src/dysh/util/calibrator.py
```diff
@@ -0,0 +1,283 @@
...
+        # Calibrator flux density in Jy.
+        snu = self.method(nu, self.coefs)
+
+        return snu
+
+
+def poly_pb(nu: Quantity, coefs: npt.ArrayLike):
+    """
+    Equation (1) in Perley & Butler 2017.
+
```
mpound, line 237: Add the equation.  e.g., :math: `log(S_{\nu} = a_0 + a_1 \log(\nu_{GHz}) + a_2 [\log(\nu_{GHz}]^2 + a^2 [\log(\nu_{GHz)]^3 + ...`

src/dysh/util/calibrator.py
```diff
@@ -0,0 +1,283 @@
...
+        self.alias = {}
+        for k, v in self.data["Objects"].items():
+            self.alias[k] = k
+            for vv in v["alias"]:
+                self.alias[vv] = k
+
+
+class Calibrator:
+    """
+    Holds calibrator details for a specific calibrator.
```
mpound, line 112: This need to be fleshed out.  The user doesn't know what 'details' means.    Explain that this class computes the spectral energy distribution using polynomial of frequency with coefficients based on literature.  etc.

src/dysh/data/calibrators.json
```diff
@@ -0,0 +1,70 @@
+{
```
mpound, line 1: Is there a valid frequency range over which the equations are valid that needs to be considered.  e.g. last column of Table 6 in PB2017

mpound (changes requested):
Solid work, a few hiccups. See comments for requested changes.

astrofle:
> for k in ct.valid_names("Ott 1994"):
>      print(k)
>      c.from_name(k)

The second error is because `from_name` uses `Perley-Butler` by default. Should this be changed to use whatever flux scale is available for that particular name if no scale is given?

src/dysh/util/calibrator.py
```diff
@@ -0,0 +1,283 @@
...
+        ------
+        ValueError
+            If `scale` is not one of the keys in the `calibrator_table`,
+            or if `name` is not one of the keys in `calibrator_table.data[scale]`.
+        """
+
+        if calibrator_table is None:
+            calibrator_table = CalibratorTable()
+
+        if scale not in calibrator_table.valid_scales():
```
mpound, line 203: Use util.core.minimum_string_match for user-friendliness?

if minimum_string_match(scale,calibrator_table.valid_scales()) is not None:

src/dysh/util/tests/test_calibrator.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestCalibratorTable:
+    def test_valid_scales(self):
+        ct = calibrator.CalibratorTable()
+        assert ct.valid_scales() == ["Perley-Butler 2017", "Ott 1994"]
+
+
+class TestCalibrator:
+    def test_from_name(self):
+        # Test some names.
```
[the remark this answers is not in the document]
astrofle: added

src/dysh/util/calibrator.py
```diff
@@ -0,0 +1,283 @@
...
+        name: str,
+        scale: Literal["Perley-Butler 2017", "Ott 1994"] = "Perley-Butler 2017",
+        calibrator_table: CalibratorTable | None = None,
+    ):
+        """
+        Create a `~dysh.util.calibrator.Calibrator` given a calibrator `name`.
+
+        Parameters
+        ----------
+        name : str
```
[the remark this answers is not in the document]
astrofle: It is now case insensitive, same for `scale`.

astrofle:
> > > for k in ct.valid_names("Ott 1994"):
> > > print(k)
> > > c.from_name(k)
> > 
> > 
> > The error is because `from_name` uses `Perley-Butler` by default. Should this be changed to use whatever flux scale is available for that particular name if no scale is given?
> 
> Hmmm...that could be confusing to the user in this (admittedly contrived) case. They silently have a calibrator on a different scale than others they might be collecting. Unless a warning is emitted and as you said yesterday warnings scare users. However, that would make it consistent with ` valid_names(None)` which is supposed to return all valid names regardless of scale. The other choices are don't default `scale` in `from_name` or add "in scale {scale}" to the error message to give the user a clue. There's no perfect solution.

I decided to default to `scale=None`. It will find the scale available, preferring PB17, and inform the user.

src/dysh/util/calibrator.py
```diff
@@ -22,35 +24,50 @@ class CalibratorTable:
...
+    "LAS" is the largest angular extent of the object in degrees.
     "Scale name" is the name of the scale, for example "Ott 1994".
+    "fluxscale" is a repeat of "Scale name".
+    "coefs" are the cofficients that define the spectral energy distribution of the object.
+    The order and the meaning of the coefficients depends on the function defined by "method".
+    "nu_min" is the minimum frequency, in GHz, where the coeffcients are valid.
+    "nu_max" is the maximum frequency, in GHz, where the coeffcients are valid.
     "method" is the method used to compute the spectral energy distribution given the list of
-    coefficients provided for each object. The method should be defined in this file.
+    coefficients provided for each object. The method should be defined in this file, calibrators.py.
```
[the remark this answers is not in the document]
astrofle: added spaces so the lines are separate

src/dysh/util/calibrator.py
```diff
@@ -22,35 +24,50 @@ class CalibratorTable:
...
 
     Parameters
     ----------
     calibrator_table_file : Path
```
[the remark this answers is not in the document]
astrofle: pathlib.Path

PR #688, opened 2025-08-12
678 performance on sdfwrite

teuben:
This PR resolved the performance on sdf.write(), which turned out to be a memory leak in
_bintable_from_rows()  in sdfitsoad.py - this leak also improved the CPU about factor 10-40 in the benchmark case, since the long copy() was not needed anymore !!

It also updated a few bench_*py code to use the re-normalized dysh_data() designations.  bench_datawrite.py was used for testing. this code had not been used until now.

In addition, the README.md file was updated, including some comments on sdmath.

Some oddities about benchmarking were noted, in particular lazy loading.


benchmark/README.md
```diff
@@ -8,54 +8,59 @@ The purpose of this  benchmarking is to:
...
 
-bench_sdmath.py: Test performance of numpy equivalent of getps.  Can be used as a baseline of sorts of the fastest we could do these operations.
+* bench_getps.py:  Test performance of getps, calibration, and averaging.
 
-bench_skel.py:  Skeleton (template) from which new benchmarks can be created.
+* bench_otf.py:  Test performance of OTF calibration
+
+* bench_sdmath.py: Test performance of numpy equivalent of getps.  Can be used as a baseline of sorts of the fastest we could do these operations.
+
+* bench_skel.py:  Skeleton (template) from which new benchmarks can be created.
```
mpound, line 24: I wonder if we shouldn't just delete bench_skel.py.   It is out of date with our other benchmarks and most likely method of creating a new benchmark is to copy an old one rather than copy the skeleton.

benchmark/README.md
```diff
@@ -92,36 +111,67 @@ if this gives permission denied, open up a root shell, and issue the command in
...
 The standard set of SDFITS files to run the benchmark on are:
 
-1. Standard positionswitch example from the notebooks - up to 4 PS on/off scans. Single fits file, 45MB - AGBT05B_047_01
-2. L-band edge data in On/Off/On mode - 8.5GB, 70 scans.   NGC2808 is extracted from this, using 9 "Track" scans in on/off/on mode. - AGBT15B_287_19
-3. ARGUS edge data in OTF mode - 1.3 GB, NGC0001 in AGBT21B_024_01 (or TBD if NGC5954 (2 strong sources) should be used - AGBT21B_024_20)
+1. Standard L-band positionswitch example from the notebooks - up to 4 PS on/off scans. Single fits file, 45MB - AGBT05B_047_01.
+   This is **example=*getps"**
+
+2. L-band edge data in On/Off/On mode - 8.5GB, 70 scans.   NGC2808 is extracted from this, using 9 "Track" scans in on/off/on mode. - AGBT15B_287_19.
+   This should be **example=*getps3"**   ???
```
mpound, line 158: What's the ??? mean ?

benchmark/README.md
teuben: I'm also in favor of deleting the skeleton.

benchmark/README.md
teuben: I probably should just remove it, since we didn't put this dataset in the example= or accept=..... I've started a fresh repo to do the HI pipeline, outside of dysh.

benchmark/README.md
```diff
@@ -131,32 +181,97 @@ times to be the same, but it all depended on if we time-averaged or not.  Times
...
+
+Few comments here:
+
+1. Disk caching clearly plays a role, especially on machines with modest memory.
+2. Python 
+
+
+```
+# writing sp and sb have similar properties
+
```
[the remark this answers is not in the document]
teuben: post.

teuben:
hmm, i thought I made it a draft....  seems to be gone.

teuben:
> You should run your benchmark in benchmark/q10stats and store the output there for future reference.

yes, I'll do that as well.  Given my comments in README comparing sdmath with getps, I suspect we could improve getps (which goes from 200ms to 400+ms if you timeaverage)

benchmark/README.md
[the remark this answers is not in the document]
mpound: We should make things clearer in README.md.    The Overall Findings section is about Q8 results.  Putting Q9 results in there (actually Q10 because we were off by one) is a bit confusing.    There should be a separate section with what Q9 fixed, which arguably I need to help write.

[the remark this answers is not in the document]
teuben: I have one more bigger update for README.md in the editor. Have a look after that commit.

benchmark/README.md
teuben: removed the 1,2,3 and referenced data by projectid, size and dysh_data alias.

teuben:
> You should run your benchmark in benchmark/q10stats and store the output there for future reference.

I'll await checking those in once you are ok with the README.

benchmark/q10stats/runbenchQ10.sh
```diff
@@ -24,26 +29,43 @@ do
...
+../bench_getps.py -d -s    -l 9 > getps_bench.log
+
+#####################
+# SDMATH
+#####################
+echo "doing SDMATH bench..."
+../bench_sdmath.py -d --loop 10 --nchan 32768 --nscan 44   --mode 0    > sdmath_bench_0.log
+../bench_sdmath.py -d --loop 10 --nchan 32768 --nscan 440  --mode 0    > sdmath_bench_1.log
+../bench_sdmath.py -d --loop 10 --nchan 32768 --nscan 4400 --mode 0    > sdmath_bench_2.log
+../bench_sdmath.py -d --loop 10 --nchan 32768 --nscan 4400 --mode 0 -p > sdmath_bench.profile
```
mpound, line 56: sdmath_bench.profile should be added to the repo so there is a record pinned to this point in time

benchmark/q10stats/runbenchQ10.sh
```diff
@@ -24,26 +29,43 @@ do
...
-echo "doing Scanblock write bench..."
-../bench_datawrite.py --writedata sb -l 10 --statslines 25 -m -p > sbwrite_bench.profile
+echo "doing SDF, SB, PS write bench..."
+../bench_datawrite.py --writedata sdf -l 10 --statslines 25       > sdfwrite_bench.log
+../bench_datawrite.py --writedata sb  -l 10 --statslines 25       > sbwrite_bench.log
+../bench_datawrite.py --writedata sp  -l 10 --statslines 25       > spwrite_bench.log
+
+../bench_datawrite.py --writedata sdf -l 10 --statslines 25 -m -p > sdfwrite_bench.profile
+../bench_datawrite.py --writedata sb  -l 10 --statslines 25 -m -p > sbwrite_bench.profile
+../bench_datawrite.py --writedata sp  -l 10 --statslines 25 -m -p > spwrite_bench.profile
```
mpound, line 68: same with *write_bench.profile

Issue #360, opened 2024-08-23
Spectrum slicing order


astrofle:
**Describe the bug**
While slicing a spectrum, there are no checks to make sure the `start:stop` interval is increasing. This can result in an empty selection.

astrofle:
The fix must accommodate the possibility of using negative values for slicing.

PR #689, opened 2025-08-14
Issue #360

astrofle:
fix: issue #360 spectrum slicing order

It should now be possible to slice using negative values, or quantities in any order. If the start is negative, it will be cast into a positive value (astropy issue). If the start and stop are not negative, they will be sorted before slicing the underlying objects.


teuben (approved):
good set of tests!

astrofle:
> I had a failing dataIO notebook.

It's likely something local. The notebooks are tested as part of the CI (only the ones under notebooks/examples) and they raised no errors.

Issue #575, opened 2025-04-30
Spectrum data and flux do not include the mask


astrofle:
**Describe the bug**
If there is a mask for a `Spectrum` object, its `data` and `flux` attributes do not include the mask. The masked channels have values of `0`. They should be masked.

**Environment**
- Dysh version 0.4.3
- Python version
- OS

PR #692, opened 2025-08-14
FIx issue #575

astrofle:
fix: issue #575

Spectrum.flux now returns a `~astropy.util.masked.MaskedQuantity` object.


mpound (approved):
It is reasonable to change the flux property but not data since data is an umasked numpy array.   Also this implementation is consistent with OneDSpectrumMixin implementation of the flux property.

PR #706, opened 2025-08-27
remove confusing message about "calibrating again"

teuben:
A confusing message
```
   Scan XXX was previously calibrated. Calibrating again.
```
could show up, getting the user worried that perhaps the signal was modified again. This is not the case, 
so the warning message (5x in scan.py) was removed.


astrofle (approved):
Looks good. Thanks Peter.

Issue #713, opened 2025-09-03
VEGAS banks out of sync results in odd INTNUM column


astrofle:
**Describe the bug**
Some SDFITS files have some integration numbers while others do not.
Originally reported by Karen O'Neil.

**How to Reproduce**
```Python
from dysh.fits import GBTOffline
sdf = GBTOffline("TRFI_050924_L1")
test = sdf.gettp(scan=2, ifnum=0, plnum=0, intnum=8, fdnum=0) # fails
test = sdf.gettp(scan=2, ifnum=1, plnum=0, intnum=8, fdnum=0) # works
```

**Environment**
- Dysh version 0.8.0
- Python version
- OS

PR #714, opened 2025-09-04
Issue 713

astrofle:
This PR fixes issue #713 by computing the integration number per FITS file instead of globally. This way, even if the time stamps (DAT-OBS) are out of sync, the integration numbers do not skip values.

### DIFF
--- a/src/dysh/fits/gbtfitsload.py
+++ b/src/dysh/fits/gbtfitsload.py
@@ -1037,7 +1037,7 @@ class GBTFITSLoad(SDFITSLoad, HistoricalBase):
         zero when the scan number changes.
         """
         if self._index is None:
-            warnings.warn("Couldn't construct integration number: index is not yet created.")  # noqa: B028
+            warnings.warn("Couldn't construct integration number: index is not yet created.", stacklevel=2)
             return
 
         # check it hasn't been constructed before.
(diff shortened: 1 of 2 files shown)


src/dysh/fits/gbtfitsload.py
```diff
@@ -1037,29 +1037,32 @@ def _construct_integration_number(self):
...
+            warnings.warn("Couldn't construct integration number: index is not yet created.", stacklevel=2)
             return
 
         # check it hasn't been constructed before.
         if "INTNUM" in self._index:
             return
         # check that GBTIDL didn't write it out at some point.
         if "INT" in self._index:
-            self._index.rename(columns={"INT": "INTNUM"}, inplace=True)  # noqa: PD002
+            self._selection = Selection(self._selection.rename(columns={"INT": "INTNUM"}))
```
mpound, line 1048: Does `inplace` not work as expected?

astrofle: It does, but ruff recommended not using it: https://towardsdatascience.com/why-you-should-probably-never-use-pandas-inplace-true-9f9f211849e4/

astrofle: I went back to the `inplace=True` version since it is faster.

```Python
%%timeit -n 100
sdf._selection.rename(columns={"INTNUM": "INT"}, inplace=True)
```
245 µs ± 17.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

```Python
%%timeit -n 100
sdf._selection = Selection(sdf._selection.rename(columns={"INT": "INTNUM"}))
```
11.4 ms ± 360 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

PR #715, opened 2025-09-04
Iplotter multiple

etsmit:
Can now do multiple IF windows on same plot with specplot.oshow(spectrum) or specplot = spectrum1.plot(oshow=[spec2,spec3,..])

### DIFF
--- a/src/dysh/plot/specplot.py
+++ b/src/dysh/plot/specplot.py
@@ -87,6 +87,7 @@ class SpectrumPlot(PlotBase):
         self._selector: InteractiveSpanSelector = None
         self._freezey = (self._plot_kwargs["ymin"] is not None) or (self._plot_kwargs["ymax"] is not None)
         self._freezex = (self._plot_kwargs["xmin"] is not None) or (self._plot_kwargs["xmax"] is not None)
+        self._oshows = []
 
     # def __call__ (see pyspeckit)
 
@@ -191,6 +192,7 @@ class SpectrumPlot(PlotBase):
         if yunit is not None:
             sf = s.flux.to(yunit)
         sf = Masked(sf, s.mask)
+
         lines = self._axis.plot(self._sa, sf, color=this_plot_kwargs["color"], lw=lw)
         self._line = lines[0]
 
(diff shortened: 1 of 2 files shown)


src/dysh/plot/specplot.py
```diff
@@ -319,19 +332,35 @@ def freexy(self):
...
+        sa = oshow_spectrum.velocity_axis_to(
+            unit=self._xunit,
+            toframe=this_plot_kwargs["vel_frame"],
+            doppler_convention=this_plot_kwargs["doppler_convention"],
+        )
+
+        lines = self._axis.plot(sa, sf, color=color, linestyle=linestyle, gid="oshow")
+
+        self.freexy()
+        self._oshows.append(lines[0])
```
astrofle, line 363: Why keep a reference to the lines?

src/dysh/plot/specplot.py
```diff
@@ -305,7 +312,13 @@ def freex(self):
         self._freezex = False
         # This line (and the other in specplot.py) will have to be addressed when we
         # implement multiple IF windows in the same plot
-        self._axis.set_xlim(self._sa.min.value, self._sa.max.value)
+        mins = []
+        maxs = []
+        for line in self._axis.lines:
```
astrofle, line 317: This would have been a place to use `self._oshows`, but you already have a way of referencing them though the `axis` object. I'd say it's best to get rid of `self._oshows`.

src/dysh/plot/specplot.py
```diff
@@ -87,6 +87,7 @@ def __init__(self, spectrum, **kwargs):
         self._selector: InteractiveSpanSelector = None
         self._freezey = (self._plot_kwargs["ymin"] is not None) or (self._plot_kwargs["ymax"] is not None)
         self._freezex = (self._plot_kwargs["xmin"] is not None) or (self._plot_kwargs["xmax"] is not None)
+        self._oshows = []
```
astrofle, line 90: See the rest of the comments about keeping this attribute.

astrofle (changes requested):
Thanks a lot Evan. Looks good.
I added a few changes in a separate PR to this branch, and have a question/comment about the `_oshows` attribute.
This needs a test. Making sure the function does not crash when used is enough IMO.
I'll create a separate issue to add support for color, linestyle et al. through the `Spectrum.plot` method.

Issue #224, opened 2024-04-02
allow boolean values when selecting SDFITS columns that are character 'T'/'F'


mpound:
**Feature description**
In methods like gettp() we have arguments sig/cal = True or False.   However, SIG and CAL in the SDFITS files are characters 'T' or 'F'.  We hide this in the method so users can deal with normal booleans.  However if one were to use the selection mechanism, one must specifiy them as chars, e.g.,   GBTFITSLoad.select(cal='T').      It would be nice to remove this discrepancy.

**Solution**
Keep a dictionary of SDFITS keywords that are boolean chars and have a simple translation method.  Basically upgrade what gettp() does to a util or fits module method.

**Additional context**

PR #727, opened 2025-09-10
Issue #224

astrofle:
Allow selection using booleans for SIG and CAL


mpound (approved):
Nice. Low hanging fruit!

PR #743, opened 2025-09-22
Bump uv version in CI

tchamberlin:
Getting fairly out of date and possibly causing problems?


astrofle:
thanks for the update

PR #757, opened 2025-09-29
750 flags not saved

mpound:
This adds
- boolean flag keyword for Scan/Scanblock.write 
- add FLAGS column to binary table output if flags is True
- set masked data to NaN on output
- tests thereof

This supercedes PR #751 which can be closed.


astrofle (approved):
The changes look good. Thanks a lot Marc.

PR #780, opened 2025-10-06
Performance and testing enhancements, and linting fixes

tchamberlin:
- **Fix exception re-raising bug in logging decorators**
- **Cache function signature derivation during logging**
- **Allow disabling history logging via DISABLE_HISTORY_LOGGING**
- **Fix linting: printf-string-formatting (UP031)**
- **Bump ruff version**
- **Fix pyupgrade linter errors**
- **Fix mutable-class-default linting**
- **target-version not needed**
- **NPY002 comment**
- **pytest-timeout**
- **Bump dep versions**
- **Improve notebook test running**
- **Convert GBO-only tests to indicate via marks**
- **logging.warn -> logging.warning**

### DIFF
--- a/notebooks/examples/tests/test_notebooks.py
+++ b/notebooks/examples/tests/test_notebooks.py
@@ -15,7 +15,7 @@ def check_notebook_execution(notebook_file):
     notebook_dir = Path(notebook_file).parent.resolve()
 
     # Open the notebook
-    with open(notebook_file, "r", encoding="utf-8") as f:
+    with open(notebook_file, encoding="utf-8") as f:
         nb = nbformat.read(f, as_version=4)
 
     # Try to execute the notebook
@@ -50,7 +50,7 @@ class TestNotebooks:
 
         # Open each notebook
         for notebook_file in self.notebook_files:
-            with open(notebook_file, "r", encoding="utf-8") as f:
+            with open(notebook_file, encoding="utf-8") as f:
                 nb = nbformat.read(f, as_version=4)
 
             top_level_headers = 0
--- a/notebooks/examples/tests/test_notebooks.py
+++ b/notebooks/examples/tests/test_notebooks.py
@@ -5,6 +5,10 @@ import nbformat
 import pytest
 from nbclient import NotebookClient
 
+# Collect notebook files at module level
+NOTEBOOK_DIR = Path("notebooks/examples/")
+NOTEBOOK_FILES = list(NOTEBOOK_DIR.glob("*.ipynb"))
+
 
 def check_notebook_execution(notebook_file):
     """Execute a given notebook and check for errors"""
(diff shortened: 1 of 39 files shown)


src/dysh/log.py
```diff
@@ -279,24 +290,31 @@ def log_call_to_result(func: Callable):
 
     """
 
+    # If logging is disabled, return the function unchanged
+    if DISABLE_HISTORY_LOGGING:
+        return func
+
+    # Cache signature at decoration time for performance
+    sig = inspect.signature(func)
```
tchamberlin, line 298: This is the key performance gain -- hoisting this out of the wrapper function means we only have to derive the signature once

pyproject.toml
```diff
@@ -151,25 +152,29 @@ extend-exclude = ["attic", "notebooks/developer", "*.ipynb"]
...
+  "I", # https://docs.astral.sh/ruff/rules/isort-i
+  "NPY", # https://docs.astral.sh/ruff/rules/numpy-specific-rules-npy
+  "PD", # https://docs.astral.sh/ruff/rules/pandas-vet-pd
+  "RUF", # https://docs.astral.sh/ruff/rules/ruff-specific-rules-ruf
+  "UP", # https://docs.astral.sh/ruff/rules/printf-string-formatting/
+  "S101", # https://docs.astral.sh/ruff/rules/assert/
+  "T10", # https://docs.astral.sh/ruff/rules/#flake8-debugger-t10
+  "PLR5501", # https://docs.astral.sh/ruff/rules/collapsible-else-if/
+  "PLR2044", # https://docs.astral.sh/ruff/rules/empty-comment/
+  "G010",
```
tchamberlin, line 168: **Note**: this PR fixes all of these retroactively, but also enforces them for all new commits via `.pre-commit-hooks.yaml`

I think `UP` is the most disruptive of these (and that URL is wrong; I need to fix -- see https://docs.astral.sh/ruff/rules/#pyupgrade-up).

For example, any commit that attempts to use the ancient `%`-style string formatting is going to fail:

```sh
$ ruff check --select UP /tmp/foo.py 
UP031 Use format specifiers instead of percent format
 --> /tmp/foo.py:3:7
  |
1 | bar="BAR"
2 | foo="FOO"
3 | print("%s%s" % (foo, bar))
  |       ^^^^^^^^^^^^^^^^^^^
  |
help: Replace with format specifiers
```

It can be auto-fixed, but it's going to require `--unsafe-fixes`:

```sh
$ ruff check --select UP /tmp/foo.py --fix --unsafe-fixes
Found 2 errors (2 fixed, 0 remaining).

$ cat /tmp/foo.py 
bar="BAR"
foo="FOO"
print(f"{foo}{bar}")
```

notebooks/examples/tests/test_notebooks.py
```diff
@@ -24,44 +28,36 @@ def check_notebook_execution(notebook_file):
...
-        # Path to notebooks.
-        self.notebook_dir = Path("notebooks/examples/")
-        # List of notebook files to test.
-        self.notebook_files = self.notebook_dir.glob("*.ipynb")
 
-    def test_notebooks_execution(self):
-        """Test that each notebook runs without errors"""
+@pytest.mark.notebooks
+@pytest.mark.slow
+@pytest.mark.parametrize("notebook_file", NOTEBOOK_FILES, ids=lambda x: x.name)
```
mpound, line 38: Nice pattern.

src/dysh/log.py
tchamberlin: I suspect there's no real world benefit here. But it is a huge relative increase (5-10x) in per-call performance at least for `log_function_call`. But unless there are any usage patterns where this is being called in a super tight loop millions of times, I don't see this impacting anything

PR #782, opened 2025-10-07
30 implement vegas channel flagging

mpound:
This PR adds the ability to flag VEGAS spurs algorithmically using the information in the FITS header.  

__New methods:__
 - `util.core.calc_vegas_spurs` calculates the location of VEGAS spurs from the FITS header variables and the given equation.  The boolean value `keep_central` determines whether the central VEGAS spur is dropped from the return list. By default it is dropped, mimicking the behavior of GBTIDL. No checking is done that the values make sense -- this is done in `flag_vegas_spurs`
 - `GBTFITSLoad.flag_vegas_spurs` - will set but not apply channel flags based given the return value of `calc_vegas_spurs`.  The following conditions must be met: 
       1. INSTRUME in the FITS header must exist and be equal to "VEGAS"
       2. The calculated spur channels must lie in the range 0 to NCHAN-1
     A warning is issues if either condition is not met.  It is assumed that VSPRVAL, VSPDELT, VSPRPIX are constant throughout the FITS file. No check is done to see if they change.  The one example where VSPRVAL changes (examples/onoff-L/data/TGBT21A_501_11.raw.vegas) has garbage values that produce fail condition 2. 

__Changes to `GBTFITSLoad`:__

A new parameter `flag_vegas` is added the the constructor with a default value True.  The parameter tells GBTFITSLoad to flag VEGAS spurs algorithmically rather than through the flag file via `flag_vegas_spurs()`.   It is independent of `skipflags`. This table describes the interaction between `skipflags` and `flag_vegas`.  If `flag_vegas=True` and flag rule with tag ending in "VEGAS_SPUR" will be ignored.
<table>
<tr><th>skipflags</th><th>flag_vegas</th><th>behavior</th></tr>
<tr><td>False</td><td>False</td><td>VEGAS and other flags are created based on the flags file</td></tr>
<tr><td>True</td><td>False</td><td>No flags are created</td></tr>
<tr><td>True</td><td>True</td><td>VEGAS flags are created based on the FITS header</td></tr>
<tr><td>False</td><td>True</td><td>VEGAS flags are created based on the FITS header.  Other flags are read in the file.</td></tr>
</table>

__New Test__:

`test_flag_vegas` in test_gbtfitsload checks that VEGAS flags created through a flag file and through the algorithm produce the same `_flagmask` array.

Other small code fixes/improvements are also implemented. 

---
__For Discussion:__
- What the correct default behavior should be.  Currently since `flag_vegas` defaults to True, users of non-VEGAS data will always get a warning when creating a `GBTFITSLoad` object, which could be confusing.  However, I don't think it should be `False` because we *want* the algorithm to supplant the file for performance reasons. 
- Is the assumption that VSP* variables do not change in an FITS file a reasonable one?

### DIFF
--- a/docs/source/getting_started/index.rst
+++ b/docs/source/getting_started/index.rst
@@ -1,8 +1,7 @@
 .. |minver| replace:: ``0.8.3``
 .. |minpatch| replace:: ``0.8.4``
 .. |pymin| replace:: 3.10
-.. |pymax| replace:: 3.12
-.. |pyupp| replace:: 3.13
+.. |pymax| replace:: 3.14
 
 ***************
 Getting Started
@@ -38,7 +37,7 @@ If you want to install your own version of ``dysh`` instead, read the following 
 Installing ``dysh``
 ===================
 
-``dysh`` requires a Python version between |pymin| and |pymax| (``dysh`` has not been tested in Python |pyupp| or above) and recent versions of
+``dysh`` requires a Python version between |pymin| and |pymax| (``dysh`` has not been tested in Python versions newer than |pymax|) and recent versions of
 `astropy <https://astropy.org>`_,
 `numpy <https://numpy.org>`_,
 `scipy <https://scipy.org>`_,
@@ -62,7 +61,7 @@ Linux and MacOS
 
     The details of how to install ``dysh`` on your own depend on what package manager you use (e.g., ``uv``, ``conda``, ``pip``).
     We strongly recommend that you install ``dysh`` in a separate Python environment, so you do not change your system Python (if you use your own anaconda3, you can of course do anything you like, though we still recommend a virtual environment).
-    It is important to note that the latest version of ``dysh`` will only work on Python versions between |pymin| and |pymax| (``dysh`` has not been tested on Python |pyupp| or above).
+    It is important to note that the latest version of ``dysh`` will only work on Python versions between |pymin| and |pymax|.
     Here we provide a few examples, which require the ability to install Python packages in your system.
 
     .. tab:: Using ``conda``
@@ -227,8 +226,7 @@ If you want to install the latest stable version of ``dysh``, then it is not nec
         conda create -n dysh-venv python==3.10
 
    This will create a new virtual environment with ``Python3.10``.
-   You can use other versions of ``Python``, but not lower than |pymin| nor above |pymax|.
-   ``dysh`` has not been tested with ``Python`` |pyupp| and above.
+   You can use other versions of ``Python``, but not lower than |pymin| nor above |pymax| (``dysh`` has not been tested on ``Python`` versions newer than |pymax|).
    You can check the ``Python`` version using the command ``python --version`` in the terminal.
 
 #. After creating the environment, activate it. Enter the following command
--- a/docs/source/whatsnew/CHANGES.rst
+++ b/docs/source/whatsnew/CHANGES.rst
@@ -2,6 +2,17 @@
 What's New?
 ***********
 
+Version 0.8.4
+=============
+
+.. _v0-8-4-bugfixes:
+
+Bug Fixes
+---------
+
+- Tested with Python 3.13 and 3.14
+- Add support for numpy 2.0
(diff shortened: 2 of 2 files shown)


mpound:
I checked the VEGAS data I have in $DYSH_DATA/example_data -- most of them have bogus values for VSP*, e.g. floating point numbers, negative values for VSPRVAL which result in very wrong channel numbers for spur locations.   The attached Table shows the analysis.   In many cases VSPRVAL changes throughout the FITS file and has weird values.   The final column "Spur(0,-1)" is the calculated first and last spur using the first-encountered values of VSP* for that file.  
```
t  = Table.read('vsptable.csv',format='ascii.ecsv')
```
to examine it.   
The python file is the script to generate the table. 

[vsptable.csv](https://github.com/user-attachments/files/22756494/vsptable.csv)
[checkvsp.py](https://github.com/user-attachments/files/22756495/checkvsp.py)

mpound:
> > Weird -- `test_vegas_flags` does not fail in my sandbox.
> 
> Are you running tests via uv on your end? If not, most likely I think would be version mismatch, since you won't be using the uv lockfile.

That could be it.  I am still using hatch.  In any event, I fixed the test.

tchamberlin:
https://hatch.pypa.io/1.12/blog/2024/05/02/hatch-v1100/#uv

I think it would pretty simple to use uv as the installer for hatch; that might be all we need to do here?

`pyproject.toml`:

```toml
[tool.hatch.envs.default]
installer = "uv"
```

mpound:
@astrofle This is now ready for review.   The vegas spurs are calculated per bintable and fitsindex, then an extra flag array `SDFITSLoad._additional_channel_flags` is set.  When flags are applied this mask is OR'd with any other channel flags in `SDFITSLoad._flagmask`.     This bypasses use of `Flag.flag` or `Flag.flag_channel` which would have been huge performance hits because they would have to be called `N_bintable*N_fitsindex*32` times.   As a result, `sdf.flags.show` will not show these flags.   It might be possible to abuse `Flag._addrow` to put entries in the "show" table but I did not want to go that route.  We could file an issue for it or leave it as is.

astrofle:
> N_fitsindex

See my review comment. This is not how this should work. You'll get values below 0 and above the number of channels in many cases, along the VEGAS spurs. SO if you avoid flagging when this happens, then you will also not flag the spurs for those rows.

src/dysh/fits/gbtfitsload.py
```diff
@@ -3449,7 +3537,7 @@ def _reload(self, force=False):
                 logger.debug("NEW MTIME:", self._mtime)
                 force = True
         if force:
-            logger.info(f"Reload {self._online}")
+            print(f"Reload {self._online}")
```
astrofle: I thought we had agreed we would use logger for the library code

mpound: Probably via a merge. Changed back to info

src/dysh/fits/tests/test_gbtfitsload.py
```diff
@@ -1656,7 +1650,7 @@ def test_vanecal(self):
         """Test for vanecal"""
 
         sdf_file = f"{self.data_dir}/AGBT21B_024_14/AGBT21B_024_14_test"
-        sdf = gbtfitsload.GBTFITSLoad(sdf_file)
+        sdf = gbtfitsload.GBTFITSLoad(sdf_file, flag_vegas=False)
```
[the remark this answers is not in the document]
mpound: Most of our test files were created without any VEGAS flags, or don't have flag files.  So there were many places I had to add flag_vegas=False to ensure that tests continue to pass.

astrofle (commented):
When loading an SDFITS with an associated flag file, ``dysh`` says:
```
INFO Flags were created from existing flag files. Use GBTFITSLoad.flags.show() to see them.
```
``GBTFITSLoad.flags.show()`` shows an empty table (as you already pointed out).
If we decide to add the VEGAS spurs to the flag table, then this message can stay as is. If not, the message should not be shown when the flag table is empty.

Personally, I think the VEGAS spurs should be shown in the flag table.

astrofle (approved):
Looks good. Thanks a lot Marc!
I think this is ready to go as soon as we have the potentially confusing flag message sorted out.

Issue #783, opened 2025-10-08
spectrum.baseline confusing use of lists vs. tuples


teuben:
**Feature description**

The documentation of spectrum.baseline() lists two examples
```
 One channel-based region: [11,51]
Two channel-based regions: [(11,51),(99,123)].
```
where it is suggested tuples need to be used if more than one region is used. Either this should be clearly stated, or we also allow [[11,51],[99,123]] .  i mean, why not?

**Solution**

Could also advertise (in  the help) that the conversion from list_of_list to list_of_tuple can be done as follows:
```
list_of_tuples = list(map(tuple, list_of_lists))
```

**Additional context**

There may be other places in dysh where tuples and lists can be interchangeably used.

Issue #779, opened 2025-10-03
spectrum.stats() does not take NaN's into account


teuben:
**Describe the bug**

The Spectrum.stats() routine uses  np.nanstd(), but for the rest it doesn't use nan resistent functions. Unless we can guarantee a mask is always properly present, this can result in a nan for the mean etc.

**How to Reproduce**

Simple user approach here, i couldn't even make it fail, why???? this is not obvious

```
ta = GBTFITSLoad(dysh_data(test="getps")).getps(ifnum=0, plnum=0, fdnum=0).timeaverage()
print(ta.flux[1])
ta.flux[1] = np.nan
print(ta.flux[1])
```

**Environment**
- Dysh version 0.8.3
- Python version 3.11
- OS  ubuntu22

teuben:
sounds like something for an FAQ.

PR #784, opened 2025-10-08
415 masks not handled in smoothing 2

teuben:
This PR solves some masking issue with smoothing, which seemed only to occur for gaussian smoothing.

 also added some new code:

1.  a new "nan" and "npt" counter is added to Spectrum.stats() - also warns if the masks are consistent, i.e. if  np.isnan(self.data).sum() is not the same as self.mask.sum() - issue #779   solved here too
2.  noting a nuisance to deal with multi-section baseline subtraction.  tuples are now required, would be nice to allow list of lists? (issue  #783 )
3.  there may be a plotter "issue" where it connects lines that should be masked.. this however goes away when spectrum.plot() is called again.
4. a new test to check number of masks in the smoothed and decimated spectrum of the 3 smoothing functions we support.

### DIFF
--- a/src/dysh/spectra/spectrum.py
+++ b/src/dysh/spectra/spectrum.py
@@ -400,7 +400,7 @@ class Spectrum(Spectrum1D, HistoricalBase):
         # note the Spectrum class has special nanXXX functions for most, but not std()
         if roll == 0:
             mean = self.mean()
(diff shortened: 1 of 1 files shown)


mpound (approved):
Nice job!

PR #794, opened 2025-10-15
618 dev docs update

mpound:
Updates to all for_developer rst files.

### DIFF
--- a/src/dysh/spectra/scan.py
+++ b/src/dysh/spectra/scan.py
@@ -2592,15 +2592,6 @@ class FSScan(ScanBase):
 
             return freq * cunit1
 
-        def do_total_power(no_cal, cal, tcal):
-            """ """
-
-        def vec_mean_tsys(on, off, tcal):
-            """
-            mean_tsys implements this, albeit only in 1D
-            """
-            pass
-
         def do_sig_ref(sig, ref, tsys, smooth=False):
             """
             smooth=True would implement smoothing the reference (or something)
@@ -2609,7 +2600,7 @@ class FSScan(ScanBase):
 
         def do_fold(sig, ref, sig_freq, ref_freq, remove_wrap=False, shift_method="fft"):
             """ """
-            chan_shift = (sig_freq[0] - ref_freq[0]) / np.abs(np.diff(sig_freq)).mean()
+            chan_shift = (ref_freq[0] - sig_freq[0]) / np.diff(sig_freq).mean()
             logger.debug(f"do_fold: {sig_freq[0]}, {ref_freq[0]},{chan_shift}")
             ref_shift = core.data_shift(ref, chan_shift, remove_wrap=remove_wrap, method=shift_method)
             # @todo weights
@@ -2622,7 +2613,6 @@ class FSScan(ScanBase):
         nspect = self._nint
         self._calibrated = np.ma.empty((nspect, self._nchan), dtype="d")
         self._calc_exposure()
-        sig_freq = self._sigcaloff[0]
         df_sig = self._sdfits.index(bintable=self._bintable_index).iloc[self._sigoffrows]
         df_ref = self._sdfits.index(bintable=self._bintable_index).iloc[self._refoffrows]
         logger.debug(f"df_sig {type(df_sig)} len(df_sig)")


docs/source/for_developers/install.rst
```diff
@@ -1,256 +1,48 @@
...
-#. Clone the repo and install hatch.
+#. Clone the repo, install `uv`, and create the virtual environment, by default in dysh/.venv:
 
     .. code-block:: bash
 
         $ git clone git@github.com:GreenBankObservatory/dysh.git
+        $ pip install uv
         $ cd dysh
-        $ pip install hatch
+        $ uv venv
```
tchamberlin: I think we should have users install uv based on uv's own docs: https://docs.astral.sh/uv/getting-started/installation/

`pip install uv` is a poor choice for most developers; this would only be necessary if you were trying to avoid a global install via the official installer, or pipx, etc. For example at GBO this would give you an error: `ERROR: Could not find an activated virtualenv (required).`

Also, `uv venv` is redundant

docs/source/for_developers/install.rst
```diff
@@ -1,256 +1,48 @@
...
 
-#. Create and activate a virtual environment with hatch and install the packages required for development.
-   The virtual environment will be created the first time; subsequent invoking ``hatch shell`` will simply load the created environment.
+#. Install dysh and its required packages. This will create a uv.lock file containing the exact dependencies.
 
     .. code-block:: bash
 
-        $ hatch shell
-        (dysh) $ pip install -r requirements_dev.txt
+       $ uv sync --all-extras --dev
```
tchamberlin: `--dev` is redundant (doubly redundant actually; even without `--all-extras`, `--dev` is implied)

docs/source/for_developers/github_integrations.rst
```diff
@@ -5,114 +5,60 @@ GitHub Integrations
...
-
-.. code-block:: bash
-
-    $ cd /path/to/dysh
-    $ hatch shell
-    (dysh) $ pip install pre-commit
-    (dysh) $ pre-commit install
-
-Now, `pre-commit` will run whenever you commit code. You can also manually run it. Consider the case where you have made a new Python file, `newfile.py`. The steps to properly commit it would look like so:
+We have several pre-commit hooks that ensure committed code follows desired standards. They can be found in `dysh/.pre-commit-config.yaml`.   pre-commit is automattically installed if you used the `dev` dependency group to install dysh (`uv sync --dev`).  `pre-commit` will run whenever you commit code.
```
tchamberlin: > automattically
automatically

tchamberlin (changes requested):
Various changes for uv-related docs

docs/source/for_developers/install.rst
mpound: Why shouldn't  `-n auto`  be used since pytest-xdist is in the dev group?

tchamberlin: That would also be fine. Keep in mind though that `-n auto` is not very intelligent -- for small numbers of tests, it will actually be slower than the default (`-n 0`, effectively), due to the overhead of spawning the runners

That's why the default args for pytest don't include `-n auto`

But ya in this case I think it would be better to include it

mpound: I left the default without it and added a comment that it is an optional argument.

docs/source/for_developers/github_integrations.rst
```diff
@@ -1,11 +1,13 @@
...
+Testing & GitHub Integrations
+********************************
 
-Actions
-=======
+Test Code
+=========
 
-Tests should be named `test_*.py` and located within a `tests` directory of their parent module. All tests are run in dysh's continuous integration suite defined in `dysh/.github/workflows/ci.yml`.
+All new features **must** have accompanying test code verifies their correctness; no PR will be accepted without it.  A jupyter notebook explaining how to use the new feature is also strongly recommended.   Bug fixes should have a regression test added to verify the issue is fixed and doesn't return.  Tests must be runnable by `pytest`.   If a test writes a file to disk, it **must** use `pytest's` `tmp_path  infrastructure <https://docs.pytest.org/en/latest/how-to/tmp_path.html>`_ to ensure the file is removed after the test.
```
astrofle: it \*\*must** use \`pytest's\` \`tmp_path infrastructure -> it \*\*must** use \`pytest\` \`tmp_path fixture

Using fixture is more accurate. The apostrophe and possessive s is a personal choice, but it felt weird seeing it there (and in code font... it really confused me).

docs/source/for_developers/doc_standards.rst
```diff
@@ -2,45 +2,45 @@
 Documentation
 *************
 
-Setting Up Sphinx Autobuilds
-============================
+Creating Documentation
+======================
 
-Here are the steps to set up Sphinx autobuilds so that you can check your documentation edits live.
+All Python classes and methods should be documented and the documentation must follow `the NumPy convention. <https://numpydoc.readthedocs.io/en/latest/format.html>`_   Some IDEs have keyboard shortcuts to insert skeleton docstrings once the method signature is created.    Use of major features should be demonstrated in an example notebook.
```
astrofle: Some IDEs have keyboard shortcuts... 
Seems weird to mention that.

docs/source/for_developers/github_integrations.rst
```diff
@@ -1,118 +1,66 @@
...
-    (dysh) $ pre-commit install
-
-Now, `pre-commit` will run whenever you commit code. You can also manually run it. Consider the case where you have made a new Python file, `newfile.py`. The steps to properly commit it would look like so:
+We have several pre-commit hooks that ensure committed code follows desired standards. They can be found in `dysh/.pre-commit-config.yaml`.   pre-commit is automatically installed if you used the `dev` dependency group to install dysh (`uv sync --dev`).  `pre-commit` will run whenever you commit code.
 
 .. code-block:: bash
 
     (dysh) $ git add newfile.py
-    (dysh) $ pre-commit
+    (dysh) $ git commit -m "adding new python file" newfile.py
```
astrofle: Can you give the filename again during the commit command?
What does that do?
(I have never done it that way)

docs/source/for_developers/github_integrations.rst
```diff
@@ -1,118 +1,66 @@
...
 
-    reformatted newfile.py
+    1 file reformatted
 
-    All done! \u2728 \U0001f370 \u2728
-    1 file reformatted.
+     ruff check...............................................................Passed
 
-Now, you have to re-add the modified file before committing it.
+If any check Failed, then a file was modified by pre-commit, so you must commit again to get those changes.
```
astrofle: If a check ?

docs/source/for_developers/github_integrations.rst
astrofle: In this case, do you still need to git add the file? Per the git documentation it seems like you don't

docs/source/for_developers/github_integrations.rst
astrofle: If a check?

docs/source/for_developers/manage_requirements.rst
```diff
@@ -2,81 +2,26 @@
...
-
-**Solution:** Compile `requirements.txt` on a non-Windows machine, like the Linux ones at GBO.
-
-Platform-dependent versions
----------------------------
-
-Some packages may require different versions for different operating systems. If you cannot resolve a dependency issue another way, there is a solution.
-
-**Solution:** Pin the versions in the `pyproject.toml` file, then re-compile the `requirements.txt`.
+Once the changed files files  are to GitHub, verify that all GitHub Actions tests pass to ensure that these requirements work for all supported versions of Python and all supported operating systems.
```
astrofle: Once the changed files are pushed to GitHub ?

astrofle (dismissed):

docs/source/for_developers/github_integrations.rst
astrofle: I read the manual :)

tchamberlin: If pre-commit modifies files, they will not be automatically staged, so you would need to re-stage them (i.e. `git add`) before attempting to commit again

edit -- ah nevermind, saw comment above. Ya you can stage them inline via this syntax

teuben:
> @teuben Please try the `uv` steps from scratch and ensure they work for you.

all working now.  I haven't appreciated when to need to source the obscure .venv/bin/activate file.

teuben:
where it says
```
$ cd docs/source
$ uv run make html
```
is it obvious that one has to do something like
```
xdg-open _build/html/index.html
```

mpound:
We can't modify uv.lock because it really should not be touched; it is completely automatic.   However, there are legitimate reasons for editing pyproject.toml, just not the dependencies section.  So I will clarify that in the instructions.

> docs say
> 
> ```
> uv add will update uv.lock and pyproject.toml; you should not edit those files directly.
> ```
> 
> ideally those files should have a comment at the top that "uv add" will control them and files should not be edited.... but then again, "uv add" might remove them? stupid.

docs/source/for_developers/github_integrations.rst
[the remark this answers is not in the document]
mpound: added

tchamberlin (approved):

PR #815, opened 2025-10-28
Quality of life improvements for dysh-lab

tchamberlin:
1. Add dysh launcher, which loads a sensible default template
2. Add all example and developer notebooks to the templates menu
3. Enable linting via jupyter lsp and basedpyright

### DIFF
--- a/src/dysh/log.py
+++ b/src/dysh/log.py
@@ -168,13 +168,41 @@ def init_logging(verbosity: int | None = None, level: int | None = None, path: P
     if path:
         config["handlers"]["dysh_instance_log_file"]["filename"] = path
         config["loggers"]["dysh"]["handlers"].append("dysh_instance_log_file")
-        logger.info(f"Log file for this instance of dysh: {path}")
 
-    # Init logging
-    logging.config.dictConfig(config)
-    logging.getLogger().setLevel(level)
-    logger.setLevel(level)
-    logger.debug(f"Logging has been set to verbosity {verbosity} / level {logging.getLevelName(level)}")
+    # Init logging - try with configured handlers, fall back if files can't be written
+    try:
+        logging.config.dictConfig(config)
+        logging.getLogger().setLevel(level)
+        logger.setLevel(level)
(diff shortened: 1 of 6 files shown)


src/dysh/shell/lab.py
```diff
@@ -1,22 +1,47 @@
...
+    shell_dir = files("dysh.shell")
 
+    # Add notebooks directory so examples subdirectory shows as templates
+    from pathlib import Path
 
-def main():
-    _args, remaining_args = parse_args()
+    import dysh
+
+    notebooks_dir = Path(dysh.__file__).parent.parent.parent / "notebooks"
```
astrofle: Instead of `Path(dysh.__file__).parent.parent.parent`, you could use (just a suggestion, in the end they do the same):
```
from dysh.util.core import get_project_root
notebooks_dir = get_project_root().parent.parent / "notebooks/examples"
```

Also, I think we should point this to "notebooks/examples", rather than "notebooks" since "notebooks/developer" contains lots of code/notebooks that no longer work.

astrofle:
Using the template still does not include logging. Is there a way to set the logging from the launcher? or does it have to be added to the template?

tchamberlin:
I think the best way would be to put the logging init in the template. I'm not sure how to get it working any other way (it would involve some hackery of the kernel initialization I think), and this way is explicit about what is going on

See latest push for implementation

image

Not sure what default level should be? Probably INFO?

mpound:
on step 2 "Open Dysh Launcher",  I would have expected 'uv run dysh-lab' to open a browser window, but I had to copy/paste the localhost url+token myself.     Is that what "Open Dysh Launcher" means?
OS: Ubuntu

tchamberlin:
Okay `dysh-lab` should now again launch browser for you. The GBO version of dysh-lab will still _not_ do this

mpound:
> Okay `dysh-lab` should now again launch browser for you. The GBO version of dysh-lab will still _not_ do this

That worked.  Why is this undesired behavior at GBO?

PR #816, opened 2025-10-28
More robust logging

tchamberlin:


tchamberlin:
- dysh no longer attempts to write to log files to the current working directory (unless explicitly asked to)
- dysh global log file now goes to the same path, regardless of working directory. It uses `platformdirs` package to determine this in a cross-platform manner (e.g. on Linux it will go to `$HOME/.local/state/dysh/log/dysh.log`). 
- Instance log now goes to wherever `--log` indicates, and is named based on pid and datetime
- If dysh cannot write to global log file, it logs a warning, but continues execution
- If dysh cannot write to the instance log file, it raises an error (since that was explicitly requested)

I've also added logging on the dysh shell init and exit to the instance log files, so they always reflect how dysh was spawned and how it exited:

```log
2025-10-31T15:59:34-0400 E 424410 dysh_instance:127 Init dysh shell via: $ .../.venv/bin/dysh -v3 --log foo.txt
2025-10-31T15:59:36-0400 I 424410 dysh_instance:160 Exiting dysh shell due to: user request
```

TODO:

- dysh-lab entry point needs to log init/exit
- Need to do a bit more testing to make sure nothing has broken
- Need to specifically make sure that re-initializing logging doesn't break anything

Issue #829, opened 2025-11-03
getfs with masked data produces spikes


astrofle:
**Describe the bug**
Calibrating data using `getfs` with flagged channels results in a calibrated spectrum which has large spikes.

**How to Reproduce**
```Python
from dysh.fits import GBTFITSLoad
fnm = "/home/scratch/psalas/support/UMD-py/test_data/TGBT22A_603_05.raw.vegas"
sdf = GBTFITSLoad(fnm, skipflags=True, flag_vegas=True)
fs = sdf.getfs(scan=12, ifnum=0, plnum=0, fdnum=0, shift_method="fft")
fs.timeaverage().plot()
```
Produces:
<img width="2000" height="1200" alt="Image" src="https://github.com/user-attachments/assets/3918c186-c6bc-4a8d-8924-bd5c30ac88e0" />

**Environment**
- Dysh version 0.9.0
- Python version
- OS

PR #831, opened 2025-11-03
Issue #829

astrofle:
Fix for issue #829 

getfs shifts fill the array with zeros before doing the fractional shift. This removes spikes in the end result.

I branched from the ap_eff and surf_err branch instead of main. The changes specific to this fix are in dysh.spectra.core.


mpound (approved):
Good job

PR #833, opened 2025-11-03
Iplotter buttons

etsmit:
Adding some buttons to the Iplotter.


astrofle:
Closing as this will be handled in a different way after the changes to the plotter

Issue #836, opened 2025-11-05
Spectrum slicing does not update Spectrum metadata


astrofle:
**Environment**
- Dysh version 0.9.0
- Python version
- OS

PR #857, opened 2025-11-07
Issue #836

astrofle:
fix: issue #836
Spectrum slicing updates the bandwdith metadata.


astrofle:
> meta["BANDWID"] = abs(meta["CDELT1"]) * (len(new_flux) - 1) # Hz
> 
> I would have written this is as len(new_flux), not the -1

Thanks @teuben . It has now been fixed. Any other issues?

teuben (approved):

Issue #858, opened 2025-11-07
Header wrong after averaging pols


jbraatz:
**Describe the bug**
After averaging polarizations, the header should show Stokes I.  It does not.  This raises concern that other header parameters are also wrong after such averaging.

**How to Reproduce**
Follow the Pol Averaging section of the Position Switched data reduction tutorial.

**Environment**
- Dysh version 0.9.0
- running on fourier using "% dysh"

PR #895, opened 2025-11-17
Averaging spectra yields correct polarization designation

etsmit:
Fix for [issue 858](https://github.com/GreenBankObservatory/dysh/issues/858).

### DIFF
--- a/src/dysh/spectra/tests/test_spectrum.py
+++ b/src/dysh/spectra/tests/test_spectrum.py
@@ -1028,13 +1028,13 @@ class TestSpectrum:
 
         # Pol designation.
         fa = f1.average(f2)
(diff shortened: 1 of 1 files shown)


src/dysh/coordinates/core.py
```diff
@@ -157,6 +158,7 @@
     -6: "YY",
     -7: "YX",
     -8: "XY",
+    0: "??",
```
mpound: I think "UNKNOWN" would be better, unless you are limiting to 2 chars.

etsmit: This is technically the label that goes in the header, so I was hoping to limit the characters, but UNKNOWN might be doable. I will test it.

etsmit:
Changes addressed. The Windows 3.13 test took 3 tries, it kept timing out on the `from_url` calls in the notebooks.

Issue #764, opened 2025-09-29
documentation for waterfall plots


astrofle:
We should provide documentation on how to use waterfall plots.

PR #897, opened 2025-11-18
764 waterfall plot documentation

etsmit:
Docs for [issue 764](https://github.com/GreenBankObservatory/dysh/issues/764)

### DIFF
--- a/src/dysh/spectra/spectrum.py
+++ b/src/dysh/spectra/spectrum.py
@@ -95,6 +95,7 @@ class Spectrum(Spectrum1D, HistoricalBase):
         else:
             self._velocity_frame = None
         self._observer = kwargs.pop("observer", None)
+        _ = kwargs.pop("psf", None)  # Hack to enable rdiv.
         Spectrum1D.__init__(self, *args, **kwargs)
         # Try making a target from meta. If it fails, don't worry about it.
         if False:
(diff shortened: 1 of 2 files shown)


src/dysh/plot/scanplot.py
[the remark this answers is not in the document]
etsmit: There aren't any overlaps.

src/dysh/plot/scanplot.py
```diff
@@ -248,3 +250,14 @@ def set_cmap(self, cmap="inferno"):
...
+    def set_norm(self, norm=None):
+        """
+        Set the norm, or zscale of the image.
+
+        Parameters
+        ----------
+        norm : str
+            norm used for the color scale. Default: "None", for linear scaling.
+        """
+        self.im.set_norm(norm)
```
astrofle: Looking at the examples in [the matplotlib documentation](https://matplotlib.org/stable/users/explain/colors/colormapnorms.html#logarithmic), it looks like this will change the values in the colorbar, in which case the colorbar label would be imprecise (there's no knowledge of if it is a linear scale or logarithmic). This should be addressed. Two options:

1. Leave as is, but document what happens when the `norm` is not linear.
2. Implement changing the colorbar label to track the current norm.

This does not need to be addressed now, but please open a new issue so we do not forget in case it is not resolved.

astrofle: Also, add a link to the matplotlib documentation on how to use `norm`.

PR #898, opened 2025-11-18
Fix plot hang issues

tchamberlin:
When working with interactive plots, dysh currently fails to manage their lifecycles properly, leading to strange/inconsistent behavior. In the extreme case, this manifests as a freeze of the GUI: plot windows can't be closed anymore and dysh has to be restarted.

This appears to all stem from lack of state management of the plots that dysh creates. This PR adds cleanup calls when plots are closed, as well as fixes some bugs in the `cid_key` event handler and SpecPlot's `onselect` call, which was incorrectly using `draw()` (blocking) instead of `draw_idle()`

This PR also adds some unit tests that demonstrate the failure case in the old code and prove that the new code doesn't similarly fail.


astrofle:
After 6 calls to `Spectrum.plot()` -- manually closing the Figure after each call --  the GUI and CLI froze 🥲.

PR #908, opened 2025-12-01
Modifying rest frequency of a Spectrum

astrofle:
New feature: expose doppler_rest property to users

This enables doing something like:
```Python
s = Spectrum.fake_spectrum()
s.doppler_rest *= 1.2
```
to modify the rest frequency of a `Spectrum`.

### DIFF
--- a/src/dysh/spectra/core.py
+++ b/src/dysh/spectra/core.py
@@ -839,7 +839,7 @@ def fft_pad(y):
     npad = newsize - nch
     nskip = npad // 2
     padded[nskip : nskip + nch] = y
-    padded[nskip : nskip + nch].mask = y.mask
+    padded.mask[nskip : nskip + nch] = y.mask
     padded[0:nskip] = padded[nskip]
     padded[nskip + nch :] = padded[nskip + nch]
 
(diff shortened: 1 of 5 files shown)


mpound:
`to()` is the method from `astropy.units.quantity.Quantity` to convert a Quantity to a conformable unit.  `s.spectral_axis.doppler_rest` is a `Quantity`.   So the error message is coming from `Quantity`.  However it is not a helpful one because `s.spectral_axis.doppler_rest.to()` will just return a new `Quantity` and not change the spectral axis at all.  Your solution `s.doppler_rest = 1e9 * u.Hz` is not correct because it creates a new attribute called `doppler_rest` which is inconsistent with that of the spectral axis:
```
s = Spectrum.fake_spectrum()
print(hasattr(s,"doppler_rest"))
s.doppler_rest = 1e9 * u.Hz
print(s.doppler_rest, s.spectral_axis.doppler_rest)
```
False
1000000000.0 Hz 1420405751.7 Hz

astrofle:
That one was tough to crack. The issue was that the code for `Spectrum.equivalencies` was setting the global equivalencies to those of a particular `Spectrum`. Changed the code to use a copy instead.

Issue #918, opened 2025-12-07
getfs does not preserve flags during fold


astrofle:
**Describe the bug**
Calibrating data with `getfs`, when using `fold=True` (default), results in a spectrum that contains the VEGAS spurs.

**Environment**
- Dysh version 0.9.1
- Python version
- OS

PR #920, opened 2025-12-09
Issue #918

astrofle:
Fix for issue #918 
`getfs` fractional shift methods now shift the mask by a fraction as well.
This is done by dilating the mask to account for the masked values being "split" between two channels.


mpound (approved):
Very nice use of binary dilation.  See my comment for changes.

PR #926, opened 2025-12-12
Issue 870 871

mpound:
Clarifies dysh installation and pre-commit usage.    Question remains if we need to remove pre-commit from devs dependency, since if people follow our `uv sync` instructions they will get a local install of pre-commit which they may not want if they prefer global install.
Also significant simplification of CONTRIBUTING.md

### DIFF
--- a/src/dysh/__init__.py
+++ b/src/dysh/__init__.py
@@ -1,6 +1,6 @@
 """Top-level package for dysh."""
 
-__version__ = "0.11.0"
+__version__ = "0.11.1"
 
 
 all = ["version"]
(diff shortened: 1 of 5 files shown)


docs/source/for_developers/github_integrations.rst
```diff
@@ -13,14 +13,51 @@ Tests must be named `test_*.py` and located within a `tests` directory of their
...
+Installing the pre-commit package
+-----------------------------------------------
+
+You can install `pre-commit` globally, so that it is active for any of your projects that have hooks, or just locally for `dysh`.  For global install, in your normal environment:
+
+.. code-block:: bash
+
+    $ uv tool install pre-commit
+
+If you prefer local to `dysh`,  the `pre-commit` package is automatically installed with `dev` dependency group when installing `dysh`:
```
astrofle: with the `dev`

PR #930, opened 2025-12-12
419 switch to fitsio v2

astrofle:


astrofle:
> @astrofle I can choose you as a reviewer since you are the author of the PR, but you should review this, too.   The windows builds are currently failing with something to do with installing fitsio, TBD.

Thanks. I'll review it early next week.

tchamberlin:
> The windows builds are currently failing with something to do with installing fitsio, TBD.

This is something I ran into with the lazy flag branch also. I think these are the choices:


1. Make fitsio optional on Windows, and add a platform marker in `pyproject.toml` so it's only required on non-Windows:
(`fitsio; sys_platform != 'win32'`). This is what I did here: https://github.com/tchamberlin/dysh/commit/07c4a02ada43b71f34f4858ff41e36c353e8b475
2. Pin setuptools for the Windows build to an older version that still has the `compiler` attribute. I haven't tested this; it might work, but would be pretty fragile.
3. Wait for a in fitsio (but issue https://github.com/esheldon/fitsio/issues/103 has been open since 2018, so this seems like a bad idea)
4. Fix fitsio somehow?

I went with option 1 because it seemed least bad. I still don't think we should spend much time on Windows support, because I expect this to nearly exclusively be run on Linux (via official GBO entry points). But this way it would technically still work on Windows, just slow (and would require a lot of RAM). But I haven't really tested any of that; treat as proof of concept only

mpound:
Status update:
I have solved the bintable issue in my sandbox, but that exposes a deeper issue.   We require RADESYS to be present  in `Spectrum` object headers in order that slicing can work.   However, if only an index file was used, RADESYS is never defined. This can happen in the `gettp` example Pedro showed, where  
```
sdfits = GBTFITSLoad("/home/astro-util/HIsurvey/Session02")
tp0 = sdfits.gettp(scan=299, ifnum=0, plnum=0, fdnum=0)
tp0.timeaverage(use_wcs=False)

  File /bigdisk/src/dysh/src/dysh/coordinates/core.py:549 in make_target
    frame=header["RADESYS"].lower(),

AttributeError: 'NoneType' object has no attribute 'lower'
```
And there are other issues with FITS headers, like trying to put NaN in them (if `use_wcs=True`)

I do not see a way to reliably construct RADESYS from the columns that are in the index file.   However, we don't want to trigger a full file read in the `GBTFITSLoad` constructor as that defeats the purpose of this branch.

I think the key is to ensure meta data required for the scan are correctly updated after `SDFITSLoad.load_full_rows` is called (e.g. in `GBTFITSLoad._load_full_rows_if_needed`).

mpound:
> Using the index file would also break Nod. We need to rewrite to part that figures out what beams are being used.

@astrofle Can you give me a minimally failing example?  All our pytests now pass with index file, but we may not be testing getnod well.

astrofle:
I'm still getting errors due to the missing RADESYS value:
```Python
path = "/home/scratch/ajschmie/training/dysh/datasets/argus/TGBT22A_603_05_vanecal.raw.vegas"
sdfits = GBTFITSLoad(path)

for i in range(16):
    print('FDnum = ', i)
    tsys = sdfits.vanecal(10, fdnum=i)
    print('   tsys = ', tsys)
```

mpound:
> I'm still getting errors due to the missing RADESYS value:
> 
> ```python
> path = "/home/scratch/ajschmie/training/dysh/datasets/argus/TGBT22A_603_05_vanecal.raw.vegas"
> sdfits = GBTFITSLoad(path)
> 
> for i in range(16):
>     print('FDnum = ', i)
>     tsys = sdfits.vanecal(10, fdnum=i)
>     print('   tsys = ', tsys)
> ```

Is the error you see: `Coordinate frame "" is not a known coordinate frame`?    That's what I get on fourier.

astrofle:
> > I'm still getting errors due to the missing RADESYS value:
> > 
> > ```python
> > path = "/home/scratch/ajschmie/training/dysh/datasets/argus/TGBT22A_603_05_vanecal.raw.vegas"
> > sdfits = GBTFITSLoad(path)
> > 
> > for i in range(16):
> >     print('FDnum = ', i)
> >     tsys = sdfits.vanecal(10, fdnum=i)
> >     print('   tsys = ', tsys)
> > ```
> 
> Is the error you see: `Coordinate frame "" is not a known coordinate frame`?    That's what I get on fourier.
> 
> 
> 

Yes, that's the error

PR #942, opened 2025-12-16
747 gbtfitsloadflagsfinal is empty

mpound:
* fixes 747 with code provided by @tchamberlin 
* adds new testdata TGBT25B_603_12.  Only a subset of the full data which is too large for the repo
* test code to ensure flag union is OR not AND

### DIFF
--- a/src/dysh/__init__.py
+++ b/src/dysh/__init__.py
@@ -1,6 +1,6 @@
 """Top-level package for dysh."""
 
-__version__ = "0.11.1"
+__version__ = "0.11.2"
 
 
 all = ["version"]


astrofle:
Thanks for the fix Marc. I think the ~200 MB in files added for the test is unnecessary, so I went ahead and removed it from the commit history in #944 . Apologies I didn't ask first.

mpound:
> Thanks for the fix Marc. I think the ~200 MB in files added for the test is unnecessary, so I went ahead and removed it from the commit history in #944 . Apologies I didn't ask first.

That's fine.  I was on the fence about it, even after paring it down significantly from the original test.

Issue #883, opened 2025-11-14
Error writing ScanBlock


astrofle:
**Describe the bug**
While trying to write a `ScanBlock` an error pops up. Originally reported by Larry Morgan.

**How to Reproduce**
```Python
# sdfm = GBTOffline("TGBT25B_608_07")
from dysh.spectra import ScanBlock
offspl0 = []
offspl1 = []
for i in range(7):
  if i == 2:
    off19pl0 = sdfm.gettp(scan=19,ifnum=1,plnum=0,fdnum=i).timeaverage()
    off58pl0 = sdfm.gettp(scan=58,ifnum=1,plnum=0,fdnum=i).timeaverage()
    offspl0.append(off19pl0.average(off58pl0))
    off19pl1 = sdfm.gettp(scan=19,ifnum=1,plnum=1,fdnum=i).timeaverage()
    off58pl1 = sdfm.gettp(scan=58,ifnum=1,plnum=1,fdnum=i).timeaverage()
    offspl1.append(off19pl1.average(off58pl1))
  else:
    off19pl0 = sdfm.gettp(scan=19,ifnum=0,plnum=0,fdnum=i).timeaverage()
    off58pl0 = sdfm.gettp(scan=58,ifnum=0,plnum=0,fdnum=i).timeaverage()
    offspl0.append(off19pl0.average(off58pl0))
    off19pl1 = sdfm.gettp(scan=19,ifnum=0,plnum=1,fdnum=i).timeaverage()
    off58pl1 = sdfm.gettp(scan=58,ifnum=0,plnum=1,fdnum=i).timeaverage()
    offspl1.append(off19pl1.average(off58pl1))
scanblock = ScanBlock()
for i in range(7):
  print(f'beam {i+1}/7')
  if i == 2:
    this_sb = sdfm.getsigref(scan=np.r_[20:54,59:78],ref=offspl0[i],ifnum=1,plnum=0,fdnum=i)
    scanblock.append(this_sb)
    this_sb = sdfm.getsigref(scan=np.r_[20:54,59:78],ref=offspl1[i],ifnum=1,plnum=1,fdnum=i)
    scanblock.append(this_sb)
  else:
    this_sb = sdfm.getsigref(scan=np.r_[20:54,59:78],ref=offspl0[i],ifnum=0,plnum=0,fdnum=i)
    scanblock.append(this_sb)
[...]

Issue #749, opened 2025-09-24
Tsys = nan not caught until calibrating


dbautista98:
**Describe the bug**
What is the bug? What did you expect to happen vs. what actually happened?

I found a dataset (TRFI_090425_L1) where there is a single ifnum for a single scan where all the channels are zero. This results in a division by zero that creates a nan when deriving the Tsys. This creates different error messages depending on how the data is retrieved. 

The first error message had me thinking that the problem was in the original sdfits data and I was looking through its header to find a float. I had to look through the code to realize that the header problem was with a new header that was being created. 

**How to Reproduce**

Dysh method 1 (warning followed by confusing error message): 
```
sdf = GBTOffline("TRFI_090425_L1")
INFO DYSH_DATA: None
Flags were created from existing flag files. Use GBTFITSLoad.flags.show() to see them.
Loaded 4 FITS files

scans = [1,2,3,4,5,6,7,8]

tpsb = sdf.gettp(scan=scans, ifnum=2, plnum=0,fdnum=0)
INFO Ignoring 1 blanked integration(s).
/home/dysh/release/latest/dysh/src/dysh/spectra/core.py:680: RuntimeWarning: invalid value encountered in scalar divide
  meanTsys = meanoff / meandiff * tcal + tcal / 2.0
INFO Ignoring 1 blanked integration(s).
INFO Ignoring 1 blanked integration(s).
INFO Ignoring 1 blanked integration(s).
INFO Ignoring 1 blanked integration(s).
INFO Ignoring 1 blanked integration(s).
INFO Ignoring 1 blanked integration(s).
INFO Ignoring 1 blanked integration(s).

tpsb[0].timeaverage()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File /home/dysh/release/latest/dysh/src/dysh/log.py:357, in log_call_to_history.<locals>.wrapper(self, *args, **kwargs)
    356 try:
--> 357     result = func(self, *args, **kwargs)
    358 except:  # remove the wrapper from the stack trace  # noqa: E722

File /home/dysh/release/latest/dysh/src/dysh/spectra/scan.py:829, in ScanBase.timeaverage(self, weights)
    828     raise Exception("You can't time average before calibration.")
--> 829 self._timeaveraged = deepcopy(self.calibrated(0))
    830 data = self._calibrated

File /home/dysh/release/latest/dysh/src/dysh/spectra/scan.py:363, in ScanBase.calibrated(self, i)
[...]

PR #943, opened 2025-12-17
Keep NaN values in ScanBlock

astrofle:
Do not remove NaN values from meta, instead, only use necessary meta for making WCS objects (issue #883). 
Also, check for NaN values in the weights array while time averaging (issue #749).

Updates tests and adds a test for `ScanBlock.write` when there are NaN values in meta.

This fixes issues #883 and #749

### DIFF
--- a/src/dysh/__init__.py
+++ b/src/dysh/__init__.py
@@ -1,6 +1,6 @@
 """Top-level package for dysh."""
 
-__version__ = "0.11.1"
+__version__ = "0.11.2"
 
 
 all = ["version"]


src/dysh/spectra/spectrum.py
```diff
@@ -1427,17 +1427,28 @@ def make_spectrum(cls, data, meta, use_wcs=True, observer_location=None, observe
...
+                    "CUNIT3",
+                    "CTYPE4",
+                    "CRVAL4",
+                    "DATE-OBS",
+                ]
+                try:
+                    wcs_meta = {k: _meta[k] for k in wcs_meta_keys}
+                except KeyError as exc:
+                    raise KeyError(f"Missing item for {exc} in meta.") from exc
+                print(wcs_meta)
```
mpound: remove print.

mpound (approved):
I verified that Larry's script runs correctly with this branch.  Don't forget to update patch in `__init__.py`. Thanks.

src/dysh/spectra/spectrum.py
astrofle: Good catch. Thanks.

PR #950, opened 2025-12-19
652 restfreq vs restfrq

mpound:
This fixes the specific issue about plotter using `Spectrum.meta["RESTFREQ"]` instead of `Spectrum.rest_value`.   It also replaces `Spectrum.doppler_rest` methods with `Spectrum.rest_value` to avoid confusion because Spectrum1D calls it rest_value but SpectralAxis calls it doppler_rest.  Also ensures that both RESTFREQ and RESTFRQ meta are set when rest_value is set.
Also a bunch of doc cleanup.


teuben:
excellent cleanup. Happy to see the new s.rest_value for my own pipeline, it works.

For clarity for future changes, maybe place the RESTFRQ and RESTFREQ settings in fake_spectrum adjacent.

PR #951, opened 2025-12-20
844 improved documentation for selection object

mpound:
* Updates selection notebook to use a GBTFITSLoad object rather than a Selection object (which users never instantiate themselves). 
* Better explanations of selection functionality and improved examples
* Adds new feature: `GBTFITSLoad.summary(selected=True)` will give a summary of `GBTFITSLoad.selection.fina`. Default is False.
* Adds UTC as valid ColDef for `summary`
* Replaces some Exceptions with more specific Exception types
* Replaces warnings.warn with logger.warning

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,10 +124,6 @@ exclude = [
   "/notebooks",
 ]
 
-[tool.hatch.build.targets.wheel.force-include]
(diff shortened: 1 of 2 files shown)


astrofle (approved):
Looks great. Thanks a lot!

PR #963, opened 2025-12-30
release-0.11.5

astrofle:
fix: remove duplicate default template from wheel

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,10 +124,6 @@ exclude = [
   "/notebooks",
 ]
 
-[tool.hatch.build.targets.wheel.force-include]
-"src/dysh/lab_templates" = "dysh/lab_templates"
-
-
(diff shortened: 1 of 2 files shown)


astrofle:
> What was the symptom that caused you to spot this?

The publish release to PyPI action failed with this error:
```
WARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Invalid distribution file. ZIP archive not accepted: Duplicate filename
         in local headers. See https://docs.pypi.org/archives for more          
         information    
```
Looking at the docs mentioned on the error they say that the wheel file should not have repeated file names. If you look at the build step, it mentions that there is a duplicate file name while creating the wheel:
https://github.com/GreenBankObservatory/dysh/actions/runs/20588830316/job/59131308269#step:5:11

PR #964, opened 2026-01-02
review example notebooks

teuben:
positionswitch draft conversion. Many more drafts coming.

### DIFF
--- a/notebooks/examples/vane.ipynb
+++ b/notebooks/examples/vane.ipynb
@@ -95,7 +95,16 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "13:44:51.777 I Resolving example=vane -> fs-Argus/data/AGBT20B_295_02.raw.vegas/AGBT20B_295_02.raw.vegas.A.fits\n"
+      "18:42:09.754 I Resolving example=vane -> fs-Argus/data/AGBT20B_295_02.raw.vegas/AGBT20B_295_02.raw.vegas.A.fits\n",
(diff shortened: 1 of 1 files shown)


astrofle:
I think that the end goal of the "final stats" section of the notebooks is better accomplished through regression tests. If you want to keep it, I won't stop you. Regardless, we should still spellcheck the notebooks, probably there's a Spyder plugin that does it. I just noted one error, but I've seen more (they might have been there from before).

Please do not make changes to stats (it is already a complicated function since its return changes depending on the use of qac), if you want to implement what you describe, do it in a separate function that uses stats.

FYI, the screenshot I added is from dysh-test on read the docs which uses Python 3.10.

teuben:
There's now a much less verbose "check_stats()" ending of the notebook that uses np.isclose() on just the rms

astrofle:
For dataIO notebook:
* remove dot at the end of section: "Calibrate a position switched scan." -> "Calibrate a position switched scan"
* after cell 10: "Here we continue with some examples how to write and read data." -> "Here we continue with some examples of how to write and read data."
* above cell 15, remove the incomplete sentence in parenthesis or end it
* fix or remove cell 17

teuben:
> For the custom_baseline notebook:
> 
> * markdown cell above cell 3: "reproducability" -> "reproducibility"

fixed.   Oddity:  I had to add plot.show() in cell 10, and there is no red "best fit" line

teuben:
> For dataIO notebook:
> 
> * remove dot at the end of section: "Calibrate a position switched scan." -> "Calibrate a position switched scan"
> * after cell 10: "Here we continue with some examples how to write and read data." -> "Here we continue with some examples of how to write and read data."
> * above cell 15, remove the incomplete sentence in parenthesis or end it
> * fix or remove cell 17

ok, done.

PR #967, opened 2026-01-05
867 user cannot flag data following tutorial

mpound:
This PR includes
- update `spectra.core.mean_tsys` to mask values that are NaN
- in `GBTFITSLoad.flag_channel`, if the inner 80% of channels are flagged, the rest of the channels will be flagged.
- update the docs and exception message to note this behavior
- regression test for 80% flagging


src/dysh/spectra/core.py
```diff
@@ -671,21 +671,22 @@ def mean_tsys(calon, caloff, tcal, mode=0, fedge=0.1, nedge=None):
     # Define the channel range once.
     chrng = slice(nedge, -(nedge - 1), 1)
 
-    # Make them doubles. Probably not worth it.
-    caloff = caloff.astype("d")
-    calon = calon.astype("d")
+    calon = np.ma.masked_array(calon, keep_mask=True, dtype=np.float64)
+    calon.mask |= np.ma.masked_where(np.isnan(calon), calon).mask
+    caloff = np.ma.masked_array(caloff, keep_mask=True, dtype=np.float64)
+    caloff.mask |= np.ma.masked_where(np.isnan(caloff), caloff).mask
```
astrofle: This seems redundant, why not just:
```Python
caloff.mask |= np.isnan(caloff)
```
?

src/dysh/spectra/core.py
```diff
@@ -671,21 +671,22 @@ def mean_tsys(calon, caloff, tcal, mode=0, fedge=0.1, nedge=None):
...
         meanTsys = meanoff / meandiff * tcal + tcal / 2.0
     else:
-        meanTsys = np.nanmean(caloff[chrng] / (calon[chrng] - caloff[chrng]))
+        meanTsys = np.ma.mean(caloff[chrng] / (calon[chrng] - caloff[chrng]))
         meanTsys = meanTsys * tcal + tcal / 2.0
 
     # meandiff can sometimes be negative, which makes Tsys negative!
     # GBTIDL also takes abs(Tsys) because it does sqrt(Tsys^2)
-    return np.abs(meanTsys)
+    return np.ma.abs(meanTsys)
```
astrofle: I understand we did this because we wanted to replicate GBTIDL, but it just seems wrong. If the computation returns a negative system temperature, then something is wrong, and that should not be swept under the rug like this. I'd suggest we drop the ``abs``.

src/dysh/spectra/scan.py
```diff
@@ -363,7 +363,7 @@ def _finish_initialization(
     ):
         if len(meta_rows) == 0:
             raise Exception(
-                f"In Scan {self.scan}, no data left to calibrate. Check blank integrations, flags, and selection."
+                f"In Scan {self.scan}, no data left to calibrate. Check blank integrations, flags, and selection. If the inner 80% of channels has been flagged, system temperature cannot be calculated."
```
astrofle: add "the": "... the system temperature ..."

src/dysh/spectra/core.py
mpound: Removed abs() but left in as comment in case we run into issues later.   All tests pass

PR #978, opened 2026-01-14
Review for PR967

astrofle:
Change how to check if inner 80% of channels are flagged. Instead of raising an exception if that is the case, log a warning.


src/dysh/util/core.py
```diff
@@ -738,52 +737,25 @@ def replace_col_astype(t: Table, colname: str, astype, fill_value):
         t[colname].mask = savemask
 
 
-def shrink_region(arr: NDArray[np.bool_], value: bool = True, percent: float = 0.2) -> NDArray[np.bool_]:
+def inner_channel_slice(nchan: int, fedge: float = 0.1) -> slice:
```
mpound: Well, gee, that's simpler!

Issue #985, opened 2026-01-16
ScanBlock write complains about different number of integrations


astrofle:
**Describe the bug**
I wanted to write several Scans, with different number of integrations, into a single SDFITS. dysh complains.

**How to Reproduce**
```Python
from dysh.fits import GBTFITSLoad
from dysh.util import get_project_testdata

fnm = get_project_testdata() / "AGBT05B_047_01/AGBT05B_047_01.raw.acs"

# Same data in both. Just different number of integrations.
sb1 = sdf.gettp(scan=51, ifnum=0, plnum=0, fdnum=0)
sb2 = sdf.gettp(scan=51, ifnum=0, plnum=0, fdnum=0, intnum=[0,1,2])

# Merge.
sb1.extend(sb2)

# Try to write. This raises an exception.
sb1.write("test_write.fits")
```

The traceback:
```Python
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 sb1.write("test_write.fits")

File [/home/scratch/psalas/python/dysh-dev/src/dysh/spectra/scan.py:1488](http://thales.gb.nrao.edu:9050/home/scratch/psalas/python/dysh-dev/src/dysh/spectra/scan.py#line=1487), in ScanBlock.write(self, fileobj, flags, output_verify, overwrite, checksum)
   1484 thisshape = np.shape(scan._calibrated)
   1485 if thisshape != datashape:
   1486     # @todo Variable length arrays? https://docs.astropy.org/en/stable/io/fits/usage/unfamiliar.html#variable-length-array-tables
   1487     # or write to separate bintables.
-> 1488     raise Exception(
   1489         f"Data shapes of scans are not equal {thisshape}!={datashape}. Can't combine Scans into single"
   1490         " BinTableHDU"
   1491     )
   1492 # check that the header keywords are the same
   1493 diff = set(scan._meta[0].keys()) - defaultkeys

Exception: Data shapes of scans are not equal (3, 32768)!=(11, 32768). Can't combine Scans into single BinTableHDU
```

**Environment**
- Dysh version 0.11.7

PR #987, opened 2026-01-16
Issue #985

astrofle:
fix: ScanBlock can write scans with different intnums (issue #985)


mpound (approved):
Good, what matters is number of channels, as you have corrected.

PR #995, opened 2026-01-22
getfs wrapping issue

astrofle:
fix: remove wrapped channels in getfs


astrofle:
@etsmit this is now ready for review

PR #1003, opened 2026-02-05
iPlotter refactor

astrofle:
Implements a BaseGUI class to handle common GUI components.
Changes behavior of replot or show.
Changes order of tests to avoid issues due to interaction of shell tests and plotting.
Adds StaticLabGUI to handle rendering in Read the Docs.
Adds a warning if trying to use the ShellGUI without the Tk event loop.

Example read the docs output: https://dysh-test.readthedocs.io/en/plotter-show-rtd/tutorials/examples/positionswitch.html#plotting

### DIFF
--- a/src/dysh/__init__.py
+++ b/src/dysh/__init__.py
@@ -7,7 +7,7 @@ warnings.filterwarnings("ignore", message="invalid value encountered in do_forma
 # ipympl multiple-inheritance MRO issue with traitlets (ipympl#488)
 warnings.filterwarnings("ignore", message="Passing unrecognized arguments to super")
 
-__version__ = "0.12.0"
+__version__ = "0.12.1"
 
 
 all = ["version"]


src/dysh/plot/labgui.py
```diff
@@ -5,35 +5,60 @@
...
-
-        if hasattr(plotbase, "_selector"):
-            self.button_clear = Button(
-                description="Clear All Regions",
-                disabled=False,
-                tooltip="Clear all regions from plot",
-            )
+        self.canvas.header_visible = False  # Remove canvas header.
+
+        self.button_clear = Button(
```
etsmit: same with `shellgui`, disable these buttons in the case of a scanplot

src/dysh/util/core.py
```diff
@@ -583,17 +583,32 @@ def grouper(iterable, n, *, incomplete="fill", fillvalue=None):
 def in_notebook() -> bool:
```
mpound: Thomas posted a newer version of `in_notebook` in slack.  Is that included here?

mpound (approved):

PR #1018, opened 2026-02-19
script plots

astrofle:
Enable halting a script to bring up a plotter

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
     hooks:
       - id: yamlfmt
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.1
+    rev: v0.15.6
     hooks:
       - id: ruff-format
       - id: ruff-check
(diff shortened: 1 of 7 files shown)


etsmit:
I'm not able to use the callbacks in this mode, for example `get_selected_regions()`. For example, in the use case of
1) cycle through 10 spectra
2) for each, plot the spectrum, interactively select good regions, then
3) auto-take out a 1st order baseline and maybe add it to a buffer for averaging or write it back to disk

This fails because the script gets paused on the plot, where I can select the regions, but in order to get to the next step in the script (`ps.baseline(..., include=ps.get_selected_regions())`), there is no more `ps._plotter` to reference.

p.s. I'm putting this down because it's a likely use case for whenever I get around to making a receiver checkout script.

teuben (approved):
I think it would still be nice to have block=True the default, as it is in matplotlib.

There are some strange blocking behaviors with mpl i the same script, as I've shown in my fake_plot.py script, but that may also be the peculiar way how I've written the interaction. See comments in that script, it's now a show stopper, more a feature?

PR #1020, opened 2026-02-20
feat: Add plotter buttons to change units

astrofle:
Adds plotter buttons and functions to change the
ordinate and abscissa axes units. Also some minor
clean up, and starts using a stylesheet to set
matplotlib rc parameters.

### DIFF
--- a/src/dysh/data/dysh.mplstyle
+++ b/src/dysh/data/dysh.mplstyle
@@ -1,0 +1,2 @@
+font.family : monospace
+axes.xmargin : 0
--- a/src/dysh/plot/labgui.py
+++ b/src/dysh/plot/labgui.py
@@ -2,11 +2,12 @@
 dysh lab (Jupyter) plotting interface.
 """
 
+import functools
 import warnings
 
 from ipympl.backend_nbagg import Canvas, FigureManager
 from IPython.display import display
-from ipywidgets import Button, HBox, VBox
+from ipywidgets import Button, Dropdown, HBox, VBox
 from matplotlib.backends.backend_agg import FigureCanvasAgg
 
 from .basegui import BaseGUI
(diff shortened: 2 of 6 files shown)


astrofle:
> Converting to Jansky units fails.
> 
> ```
> from dysh.spectra import Spectrum
> Spectrum.fake_spectrum().plot()
> [pull down to Jy or mJy]
> astropy.units.errors.UnitConversionError: 'K' (temperature) and 'mJy' (spectral flux density) are not convertible
> ```
> 
> Should convert to Jy/beam using astropy equivalency
> 
> ```
> equiv = u.brightness_temperature(freq, beam_area)
> flux.to(newunits,equivalencies=equiv)
> ```

I've changed the behavior so that it only allows changing between K and mK or Jy and mJy. Without a dish diameter or beam size we cannot properly set the equivalency between Jy and K (or at least I couldn't get it to produce accurate numbers).

mpound:
> > Converting to Jansky units fails.
> > ```
> > from dysh.spectra import Spectrum
> > Spectrum.fake_spectrum().plot()
> > [pull down to Jy or mJy]
> > astropy.units.errors.UnitConversionError: 'K' (temperature) and 'mJy' (spectral flux density) are not convertible
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should convert to Jy/beam using astropy equivalency
> > ```
> > equiv = u.brightness_temperature(freq, beam_area)
> > flux.to(newunits,equivalencies=equiv)
> > ```
> 
> I've changed the behavior so that it only allows changing between K and mK or Jy and mJy. Without a dish diameter or beam size we cannot properly set the equivalency between Jy and K (or at least I couldn't get it to produce accurate numbers).

We can add an RFE for the `Observatory` class to implement a `beam(freq:Quantity)->Quantity`  and/or `dish_diameter->Quantity` property .   The TELESCOP in the spectrum header would allow to look this up.

PR #1027, opened 2026-03-02
Lazy flags

mpound:
- Adds lazy flag initialization to GBT/SDFITSLoad in order to reduce memory allocation and start up time
- Adds "by-rows" (chunked) writing to GBT/SDFITSLoad to avoid creating the full dense numpy array for flags at write time
- Does conditional import of FITSIO to avoid windows CI failures


astrofle:
> > Looks good. Thanks. I'll note that directly modifying the `.mask` attribute of a `Scan` is properly picked up by plotting, but not by writing. Should be documented IMO.
> 
> Do you mean Spectrum, not Scan?   If the user changes `Scan._calibrated.mask` it would get written properly. 

I see. I forgot to use the proper attribute. No need to do anything then.

PR #1044, opened 2026-03-11
getsigref fixes - part 2

teuben:
forgotten change back to the old comparison pre-hack


astrofle:
Remove the commented out lines and it's good to go

PR #1048, opened 2026-03-13
Add flag_vegas to calseq and getvane

astrofle:
Also, removes the "TSYS" message from ``calseq``.


mpound (approved):
Thanks for catching those.  Can you also remove the old method now named `_flag_vegas_spurs`?

PR #1052, opened 2026-03-14
No spur flags non vegas

astrofle:
Avoid generating VEGAS spurs for non-VEGAS backends, and reverts some typos Pedro made in the tests (thanks to Marc for pointing them out).

### DIFF
--- a/src/dysh/fits/gbtfitsload.py
+++ b/src/dysh/fits/gbtfitsload.py
@@ -1366,6 +1366,23 @@ class GBTFITSLoad(SDFITSLoad, HistoricalBase):
         """
         self._flag.flag_channel(tag=tag, channel=channel)
 
+    @property
+    def backend(self) -> str:
+        """Return the backend value or 'unknown' if it can't be currently determined.
+
+        **Note:** Some SDFITS files do not have the BACKEND or INSTRUME keywords in the
+        primary header.   If the FITS metadata were loaded from an index file rather than the
+        SDFITS binary table, the return string could be 'unknown' in this case even if
+        the binary table properly reflects the BACKEND value.
+        """
+        if "INSTRUME" in self._selection:
+            backend = str(next(iter(set(self["INSTRUME"])))).upper()
+        if "BACKEND" in self._selection:
+            backend = str(next(iter(set(self["BACKEND"])))).upper()
+        else:
+            backend = "unknown"
+        return backend
+
     def is_vegas(self):
         """Check if these data appear to use the VEGAS backend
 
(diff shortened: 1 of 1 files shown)


mpound:
I can't find the putative VEGAS data that did not have INSTRUME or BACKEND in the header, but let's say they exists.   If INSTRUME or BACKEND are not in the primary header ``GBTFITSLoad.is_vegas()`` will return False.  If INSTRUMENT or BACKEND is not "VEGAS" ``GBTFITSLoad.is_vegas()`` will return False.  

We have these cases:
1) VEGAS data that have proper INSTRUME or BACKEND in the primary header and
       a) have valid VSP columns in the FITS table
       b) have garbage VSP columns in the FITS table
2) VEGAS data that do not have proper INSTRUME or BACKEND in the primary header but do have BACKEND in the FITS data and
      a) have valid VSP columns in the FITS table
      b) have garbage VSP columns in the FITS table
3) Non-VEGAS data that have INSTRUME or BACKEND in the primary header that is not "VEGAS" and 
      a) do not have VSP columns in the FITS table
      b) have garbage VSP columns in the FITS table (should not exist but let's list it)
 4) Non-VEGAS data that have neither INSTRUME or BACKEND  in the primary header nor VSP columns in the FITS table

- Case 1a is dealt with in the current code in `main` branch
- Case 1b is dealt with by ``util.calc_vegas_spurs`` (because I had examples of this when I wrote it) -- **But I will verify**
- Case 2a is now dealt with by the new flag_vegas_spurs which handles partial loading. 
- Case 2b is dealt with  by ``util.calc_vegas_spurs`` 
- Case 3a:  I would have thought that the current code in `main` handled this, but perhaps not.
- Case 3b should be dealt with by ``util.calc_vegas_spurs`` (same caveat as 1b)
- Case 4 is dealt with by current `main` code 

I think the structure of ``flag_vegas_spurs`` should be:
```
if self.is_vegas() == False:
    - Warn that backend could not be determined or isn't VEGAS
    - Enter the loop that loads VSP* for rows in question.  
       - If they are not there, the method will exit from a KeyError with a warning.   
       - If they are there, its VEGAS data, let ``calc_vegas_spurs`` do its thing.
```     
We perhaps should try to separate out backend could not be determined from backend isn't vegas.  I.e. 
```
def backend(self) -> str:
      if INSTRUMENT or BACKEND not present reutrn "unknown" otherwise return the value
```
Then ``flag_vegas_spurs only`` tries to flag if backend() != "unknown"

I will dig and then make a PR against this PR later (might not be until tomorrow).

Issue #474, opened 2025-02-05
header display for Spectrum objects


astrofle:
**Feature description**
Display the header for a `Spectrum` object in user readable format.

**Solution**
Add a header function to `Spectrum`.

**Additional context**
This is similar to the functionality provided by the [`header` procedure in GBTIDL](https://www.gb.nrao.edu/GBT/DA/gbtidl/release/user/guide/header.html).

mpound:
Is this sufficient?

```
@property
def header(self):
    return self.meta
```

astrofle:
Not quite. We want something that prints information a user can understand, hopefully, without a lot of background (like having to check what each key means). GBTIDL does the following:

```IDL
filein,"/home/dysh/example_data/rxco-W/data/TSCAL_220105_W.raw.vegas/TSCAL_220105_W.raw.vegas.A.fits"
gettp,23
header
```
Will print
```
--------------------------------------------------------------------------------
Proj: TSCAL_220105_W   Src   : 2253+1608                   Obs : Larry Morgan   

Scan:        23        RADec :  22 53 57.6  +16 08 52     Fsky:  76.995353 GHz
Int :         0        Eqnx  :  2000.0                    Frst:  77.000000 GHz
Pol :        YY        V     :  0.0          RADI-LSR     BW  :   1.500    GHz
IF  :         0        AzEl  :  201.643      66.529       delF:   1.465    MHz
Feed:         1        Gal   :   86.110     -38.184       Exp :  44.6      s
Proc:     CALSEQ       UT    : +21 47 44.0   2022-01-05   Tcal:    1.00    K
Seqn:         1        LST/HA: +23 30 14.3   0.60         Tsys:    1.00    K
--------------------------------------------------------------------------------
```

PR #1074, opened 2026-03-25
474 spectrum header

etsmit:
Repost for merging into `release-1.1`.

Feature for https://github.com/GreenBankObservatory/dysh/issues/474.

We should address the intnum at some point, and this could definitely be refactored for a ScanBlock header(), but I didn't want to think about how the information about N scans could be distilled. This is the simplest version of a header, just for spectrum objects.

### DIFF
--- a/src/dysh/plot/core.py
+++ b/src/dysh/plot/core.py
@@ -38,4 +38,3 @@ def parse_html(s):
     s = s.replace(" (TopModel)", "")
 
     return s
-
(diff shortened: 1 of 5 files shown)


astrofle (changes requested):
See also the comments Marc left in the closed PR (you can change the base branch with the edit button)

astrofle (approved):
Thanks for doing this Evan.
(The changes I had requested could be implemented in the future.)

PR #1103, opened 2026-04-17
Fix LazyFlagArray memmap temp file cleanup (#1089)

### LINKED ISSUE #1089: LazyFlags not cleaning up after exit
If using ``np.memmap`` for storing ``LazyFlags``, the created file stays on disk even after ``dysh`` exits. This can end up filling the ``/tmp`` directory of the host. When no more memory is available, there is no error message being issued (in dysh-lab at least), the kernel just dies.

https://github.com/GreenBankObservatory/dysh/blob/34c76c0731b6d1f28306db2d28895c993c67d2f0/src/dysh/fits/lazyflag.py#L329-L333

mpound:
## Summary
- `LazyFlagArray.to_dense()` created memmap-backed temp files with `delete=False` but never cleaned them up, gradually filling `/tmp` and eventually crashing Jupyter kernels
- Track temp file paths on the instance and clean them up via `cleanup()`, `__del__`, and an `atexit` handler (using `weakref` to avoid preventing GC)
- Added tests for cleanup, deletion, and idempotent cleanup calls

Fixes #1089

## Test plan
- [x] All existing `test_lazyflag.py` tests pass (35/35)
- [x] New `TestLazyFlagCleanup` tests verify temp files are created and removed

### DIFF
--- a/src/dysh/fits/lazyflag.py
+++ b/src/dysh/fits/lazyflag.py
@@ -9,6 +9,7 @@ from ~40GB to a few MB (proportional to flagged rows only).
 import atexit
 import gc
 import os
+import shutil
 import tempfile
 import weakref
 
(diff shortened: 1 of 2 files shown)


mpound:
The new test is failing under windows.  I am investigating

astrofle (approved):
Thanks. It seems to work fine when restarting a Jupyter kernel (it removes the files created for that session). This does not address the issue of not raising any warnings/errors when there's not enough disk space.

mpound:
Warning added and a test for that, too.

PR #1104, opened 2026-04-17
Fix issue #1091: fitsio writer now persists in-memory DATA mutations

### LINKED ISSUE #1091: fitsio writer does not write updated data column
**Describe the bug**
If the "DATA" column is updated in a ``GBTFITSLoad`` object, writing a file with ``HAS_FITSIO=True`` writes the in disk "DATA" values.

**How to Reproduce**
```Python
import numpy as np
from dysh.fits import gbtfitsload

fnm = get_project_testdata() / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas/AGBT18B_354_03.raw.vegas.A.fits"

sdf = gbtfitsload.GBTFITSLoad(fnm)
sdf["DATA"] = np.ones(sdf["DATA"].shape)

sdf.write("example.fits", overwrite=True)

sdf2 = gbtfitsload.GBTFITSLoad("example.fits")
sdf2["DATA"] # Not ones.

gbtfitsload.HAS_FITSIO = False
sdf.write("example.fits", overwrite=True)
sdf2 = gbtfitsload.GBTFITSLoad("example.fits")
sdf2["DATA"] # Ones.
```

**Environment**
- Dysh version 1.0.1

mpound:
## Summary
- `_build_chunk_recarray` explicitly skipped the DATA column when overlaying in-memory mutations, on the false assumption that DATA is never mutated. `sdf["DATA"] = ...` does mutate it, so writes via the fitsio path silently lost those updates.
- Removed the DATA skip so the fitsio path now matches the astropy path's behavior.
- Added a regression test parametrized over both write paths via a monkeypatched `HAS_FITSIO`.

Fixes #1091.

## Test plan
- [x] New test `test_write_mutated_data[True]` (fitsio path) passes
- [x] New test `test_write_mutated_data[False]` (astropy path) passes
- [x] Full `src/dysh/fits/tests/test_gbtfitsload.py` suite: 86 passed

### DIFF
--- a/src/dysh/fits/gbtfitsload.py
+++ b/src/dysh/fits/gbtfitsload.py
@@ -214,15 +214,12 @@ def _build_chunk_recarray(sdf, bintable_idx, chunk_rows, col_names, include_flag
         for name in mem_only_cols:
             chunk[name] = bt_data[name][chunk_rows]
 
-    # If astropy .data was previously loaded, overlay scalar disk columns from memory
-    # to capture any mutations from sdf["COLNAME"] = value.
-    # DATA is never mutated so we always keep the fitsio version.
+    # If astropy .data was previously loaded, overlay disk columns from memory
+    # to capture any mutations from sdf["COLNAME"] = value (including DATA).
     if data_loaded:
         if not mem_only_cols:
             bt_data = bt.data
         for name in disk_cols:
-            if name == "DATA":
-                continue
             chunk[name] = bt_data[name][chunk_rows]
 
     if include_flags:
--- a/src/dysh/fits/tests/test_gbtfitsload.py
+++ b/src/dysh/fits/tests/test_gbtfitsload.py
@@ -1048,6 +1048,26 @@ class TestGBTFITSLoad:
             f = util.get_project_testdata() / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas/"
             g = gbtfitsload.GBTFITSLoad(f)
 
+    @pytest.mark.parametrize("has_fitsio", [True, False])
+    def test_write_mutated_data(self, tmp_path, monkeypatch, has_fitsio):
+        """Regression test for issue #1091: writing must persist in-memory DATA mutations
+        on both the fitsio and astropy code paths."""
+        if has_fitsio and not HAS_FITSIO:
+            pytest.skip("fitsio not installed")
+        monkeypatch.setattr(gbtfitsload, "HAS_FITSIO", has_fitsio)
+
+        f = util.get_project_testdata() / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas/AGBT18B_354_03.raw.vegas.A.fits"
+        sdf = gbtfitsload.GBTFITSLoad(f, index_file_threshold=100000000)
+        new_data = np.ones(sdf["DATA"].shape, dtype=sdf["DATA"].dtype)
+        with pytest.warns(UserWarning):
+            sdf["DATA"] = new_data
+
+        out = tmp_path / f"mutated_data_{has_fitsio}.fits"
+        sdf.write(out, overwrite=True, flags=False)
+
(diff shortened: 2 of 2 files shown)


mpound:
# context to understand the changes
# Issue #1091 — fitsio writer does not write updated DATA column

## Context

**Bug summary:** When `HAS_FITSIO=True`, `GBTFITSLoad.write()` silently ignores in-memory mutations to the `DATA` column and writes the original on-disk values. The astropy path (`HAS_FITSIO=False`) works correctly.

**Reproducer (from issue):**
```python
sdf = gbtfitsload.GBTFITSLoad(fnm)
sdf["DATA"] = np.ones(sdf["DATA"].shape)
sdf.write("example.fits", overwrite=True)
sdf2 = gbtfitsload.GBTFITSLoad("example.fits")
sdf2["DATA"]  # should be ones, but isn't when HAS_FITSIO=True
```

**Why this matters:** Users who update DATA (e.g., calibrated spectra, baseline-subtracted values) and write out the result currently get incorrect files. The bug is silent — no warning or error. This is a data-integrity bug.

## Root cause

`_build_chunk_recarray()` in `src/dysh/fits/gbtfitsload.py` (lines 138–231) reads all columns from disk via fitsio, then overlays any in-memory mutations for scalar columns. The overlay loop at lines 220–226 **explicitly skips `DATA`**:

```python
if data_loaded:
    if not mem_only_cols:
        bt_data = bt.data
    for name in disk_cols:
        if name == "DATA":
            continue                          # ← bug: skips mutated DATA
        chunk[name] = bt_data[name][chunk_rows]
```

The comment on line 219 ("DATA is never mutated") is wrong. `SDFITSLoad.__setitem__` (sdfitsload.py:1476) → `_update_binary_table_column` (sdfitsload.py:1323) mutates `self._bintable[b].data["DATA"]` directly (line 1361).

The astropy path writes correctly because `_bintable_from_rows()` pulls rows straight from `self._bintable[b].data[rows]`, which reflects the mutation.

## Fix

Remove the `if name == "DATA": continue` skip in `_build_chunk_recarray`. When `data_loaded` is `True`, the in-memory `bt.data["DATA"]` is the source of truth — whether identical to disk (no mutation) or mutated, copying from memory is correct.

- **Memory cost:** zero. `data_loaded=True` means astropy has already loaded the full `FITS_rec`, including DATA. The fix actually shrinks the fitsio read slightly (DATA drops out of `disk_cols`).
- **Correctness:** covers the mutated case; equivalent behavior in the unmutated case.
- **Streaming path preserved:** when `data_loaded=False` (user never accessed DATA), DATA still streams from disk via fitsio — unchanged.

### Files to modify

- `src/dysh/fits/gbtfitsload.py` (lines 217–226) — remove the DATA skip, correct the misleading comment.

### Test to add

Add a regression test in `src/dysh/fits/tests/test_gbtfitsload.py` (patterns at `test_set_item`, lines 925–1026). The test must exercise **both paths** by monkeypatching `dysh.fits.gbtfitsload.HAS_FITSIO` (parametrized with `pytest.mark.parametrize("has_fitsio", [True, False])`):

1. Load a small test FITS file (e.g., `AGBT18B_354_03.raw.vegas.A.fits` — used in `test_data_access`, test_sdfitsload.py:285).
2. Assign `sdf["DATA"] = np.ones(...)`.
3. Write to `tmp_path`.
4. Re-load and assert the re-loaded DATA equals the assigned values.

Use `pytest`'s `monkeypatch` fixture and `tmp_path` fixture. Expect a `UserWarning` on the assignment (existing behavior).

## Verification

1. Run the issue reproducer manually — confirm `sdf2["DATA"]` is ones in both paths.
2. `uv run pytest src/dysh/fits/tests/test_gbtfitsload.py` — new test passes.
3. `uv run pytest` — full suite still green.

astrofle (approved):
Thanks

PR #1109, opened 2026-04-29
Fix log decorators dropping kwargs and stripping string quotes (#1108)

### LINKED ISSUE #1108: log decorators don't record all arguments
**Describe the bug**
The decorator ``@log_call_to_history`` is supposed to log a method's calling arguments to the ``history`` attribute  of the calling object.   This does not always happen.  It seems to be related to default or required arguments.    Also string parameters lose their quotation marks when written to history.

**How to Reproduce**
The 2nd assertion below fails. 
```
import time
s = Spectrum.fake_spectrum()
q1=s.smooth('gaussian',5)
assert("smooth(gaussian,5,)") in s.history[-1]
# Sleep for a second to ensure the time tags for each operation are distinct, which affects
# sorting of the history list
time.sleep(1)  
q2=s.smooth(kernel='gaussian',width=5)
print(s.history[-2:])  # no arguments in second smooth call!
assert("smooth(kernel='gaussian',width=5)") in s.history[-1]
```

**Environment**
- Dysh version 1.0.1
- Python version 3.13
- OS Ubuntu 24

mpound:
## Summary
- Fixes #1108. The history-logging decorators (`@log_call_to_history`, `@log_call_to_result`, `@log_function_call`) gated kwargs logging on `"kwargs" in sig.parameters`, which is only true when the decorated function literally declares `**kwargs`. Methods like `Spectrum.smooth(kernel=..., width=...)` had their kwargs silently dropped from history. The check is removed; caller-supplied kwargs are always logged.
- `format_dysh_log_record` and `log_function_call` now use `repr()` for arg/kwarg values, so string arguments retain their quotes (`'gaussian'` instead of `gaussian`).
- Drops the now-unused `inspect` import and signature caches.

## Test plan
- [x] New `src/dysh/tests/test_log.py` covers positional, keyword, and result-side keyword call forms from the issue (3 passed).
- [x] `uv run pytest src/dysh/spectra/tests/test_spectrum.py` — 44 passed.
- [x] `uv run pytest src/dysh/spectra/tests/test_scan.py` — 42 passed, 1 skipped (pre-existing).
- [x] `uv run ruff check src/dysh/log.py src/dysh/tests/test_log.py` — clean.

### DIFF
--- a/src/dysh/log.py
+++ b/src/dysh/log.py
@@ -1,4 +1,3 @@
-import inspect
 import logging
 import logging.config
 import os
@@ -214,9 +213,6 @@ def log_function_call(log_level: str = "info"):
         except KeyError:
             raise Exception(f"Log level {log_level} unrecognized. Must be one of {list(logging._nameToLevel.keys())}.")  # noqa: B904
 
-        # Cache signature at decoration time for performance
-        sig = inspect.signature(func)
-
         @wraps(func)
         def func_wrapper(*args, **kwargs):
             try:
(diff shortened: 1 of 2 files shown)


teuben (approved):
nice, has not seen the !r modifier in f-strings

mpound:
> nice, has not seen the !r modifier in f-strings

yeah, new to me, too. thanks

PR #1110, opened 2026-04-29
1100 defining regions for basleine fitting doesnt work

mpound:
This PR promotes some hidden keywords in `spectra.core.baseline` to method parameters in both `spectra.core.baseline` and  `Spectrum.baseline`.   The keywords control what happens when baselining a spectrum that already has exclude regions set.  

```
exclude_action : str
How to combine the input exclude region with any existing exclude regions in the input Spectrum. Options are

- “replace” : replace the Spectrum.exclude_regions list with the input region

- “append” : append the input region to the Spectrum.exclude_regions list.

- None : Use the existing Spectrum.exclude_regions, ignoring the input region.
```

Also ``remove``, ``model``, ``fitter`` , ``clip_exclude``, and ``exclude_region_upper_bounds`` are promoted.

The behavior is documented with examples in the Quick Start guide under *Baseline Removal - Using Exclude Regions*.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -153,6 +153,41 @@ Plot again, but focus on the line-free regions:
 
 .. image:: files/spectrum_plot_bsub.png
 
+
+Using Exclude Regions
+--------------------------------
+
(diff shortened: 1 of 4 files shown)


teuben (approved):
sensible patch.  There was a comment in test_spectrum that needed an update once 1109 was merged. That can be done now too.

PR #1120, opened 2026-06-08
Speedup rtd

mpound:
As I noted in slack, readthedocs no longer builds because rendering our notebooks takes too long and the build times out.   
Claude identified a few things that could speed it up

  - Save local copies of intersphinx objects. The would elimate 7 remote network fetches
  eliminated per build.   I tried this is there was not a noticeable improvement, the RTD build still timed out.
  
  For the remaining bottleneck — notebook re-execution (23 notebooks, all with empty 
  outputs):

  - Option A: Commit jupyter_cache — run all notebooks locally, commit the populated jupyter_cache/ directory.  RTD reads the cache and skips execution. Quickest to implement, no workflow change. Downside: cache is binary/large, must be re-committed when notebooks change.
  - Option B: Save outputs into notebooks + skip execution on RTD — change workflow so notebooks are always committed with saved outputs.   Add this to conf.py:
 ```
    if os.environ.get("READTHEDOCS"):
        nb_execution_mode = "off"
```
  - RTD uses saved outputs, local builds still execute via cache. Cleaner long-term but requires  a discipline change — developers must run and save notebooks before committing.

  Not viable:
  - nb_execution_mode = "off" unconditionally — all notebooks have empty outputs, docs would be blank.
  - nb_execution_mode = "auto" — no improvement since all notebooks have empty cells anyway.
  
  Recommendation: Option A for immediate relief on RTD timeouts; Option B if the team wants to standardize on saving outputs. The two aren't mutually exclusive — you could do A now and migrate to B over time.

----

This branch implements option A.   Option B would not work for us, becasue we deliberately set READTHEDOCS to True so that dysh uses StaticLabGUI.    Implementing Option A fixes the failed RTD builds.


astrofle:
Some comments specific to this solution:
* Fetching inv files for intersphinx is not a significant source of run time (see [this branch](https://app.readthedocs.org/projects/dysh-test/builds/33051479/), 2m33s, vs [a branch with intersphinx fetches but without notebook execution](https://app.readthedocs.org/projects/dysh-test/builds/33076556/), 2m10s). I mention this because keeping a local copy of the inv files exposes us to get out of sync with the external repos.
* Updating a notebook requires an extra step (updating `jupyter_cache`). We'd need to automate this, otherwise it is likely we'd forget to do it (option B would also require some automation of the notebook execution, or more discipline).

astrofle:
As an side, Clause is wrong, the notebooks are stored with outputs in GitHub. Its recommendations are based on inacurate information.

tchamberlin:
This seems like the wrong direction to me; like Pedro said we already have all the outputs in git, so it seems like there should be a way to just re-use those rather than add another intermediate product to the repo. I will poke around a bit

mpound:
> Some comments specific to this solution:
> 
> * Fetching inv files for intersphinx is not a significant source of run time (see [this branch](https://app.readthedocs.org/projects/dysh-test/builds/33051479/), 2m33s, vs [a branch with intersphinx fetches but without notebook execution](https://app.readthedocs.org/projects/dysh-test/builds/33076556/), 2m10s). I mention this because keeping a local copy of the inv files exposes us to get out of sync with the external repos.
> * Updating a notebook requires an extra step (updating `jupyter_cache`). We'd need to automate this, otherwise it is likely we'd forget to do it (option B would also require some automation of the notebook execution, or more discipline).

I agree that intersphinx local copies are not worth it.  I actually removed that from this PR.

mpound:
> This seems like the wrong direction to me; like Pedro said we already have all the outputs in git, so it seems like there should be a way to just re-use those rather than add another intermediate product to the repo. I will poke around a bit

I didn't like putting `jupyter_cache` in the repo, but it was the only thing that brought the RTD build time to below the timeout.   This is a showstopper if we cannot get the RTD builds below 15m. 
FYI, There is a lot of outdated info on the web about how to fix long build times.

mpound:
We are going with option B instead.

PR #1121, opened 2026-06-08
1117 getspec fails

mpound:
The underlying issue was that loading the index file (for speed) means certain SDFITS columns required my `Spectrum.make_spectrum` aren't present.  The solution is to call `GBTFITSLoad._load_full_rows_if_needed` before calling trying to create the `Spectrum`.

Added a regression test, too.


astrofle (approved):
Thanks a lot

PR #1122, opened 2026-06-09
Fix Splatalogue CI failures (#1062)

### LINKED ISSUE #1062: Splatalogue still breaking unit tests
#1050 doesn't seem to have fixed this. See e.g. https://github.com/GreenBankObservatory/dysh/actions/runs/23301057945/job/67761966079

I think I have a fix; will open PR later today

To clarify though, we do want to keep _trying_ to run Splatalogue-related tests in CI?

mpound:
## Summary

- Fix `ALLOWED_ERROR_NAMES` in notebook tests to use bare class names (e.g. `"ConnectTimeout"`) that actually match the `ename` field nbclient checks — the previous fully-qualified names (e.g. `"requests.exceptions.ConnectTimeout"`) never matched anything
- Add `mock_splatalogue_query` fixture to `conftest.py` so the splatalogue remote code path can be tested without network access
- Add `TestSearchMocked` class with 5 tests covering dysh's own transformation layer: `orderedfreq`→`rest_frequency` rename, `obs_frequency` computation, redshift math, `intintensity` float coercion, and column filtering
- Add `"NameError"` to `ALLOWED_ERROR_NAMES` to handle cascading failures: when a network-allowed error (e.g. `ConnectTimeout`) prevents a variable from being assigned, subsequent cells that reference it raise `NameError`

## Test plan

- [x] `uv run pytest src/dysh/line/tests/ -m "not requires_internet"` — all 8 tests pass
- [ ] Confirm CI no longer fails when Splatalogue is unreachable (notebook `ConnectTimeout` now allowed)
- [x] Optionally run `uv run pytest src/dysh/line/tests/ -m requires_internet` with network access to confirm remote tests still work

Closes #1062

### DIFF
--- a/src/dysh/util/download.py
+++ b/src/dysh/util/download.py
@@ -34,7 +34,7 @@ def from_url(url, path=Path(".")):
 
     try:
         # Make the HTTPX client
-        client = httpx.Client(follow_redirects=True, timeout=30.0)
+        client = httpx.Client(follow_redirects=True)
 
         with client.stream("GET", url) as resp:
             # Get the filename from the URL
--- a/src/dysh/util/download.py
+++ b/src/dysh/util/download.py
@@ -34,7 +34,7 @@ def from_url(url, path=Path(".")):
 
     try:
         # Make the HTTPX client
-        client = httpx.Client(follow_redirects=True)
+        client = httpx.Client(follow_redirects=True, timeout=30.0)
 
         with client.stream("GET", url) as resp:
             # Get the filename from the URL
--- a/notebooks/examples/tests/test_notebooks.py
+++ b/notebooks/examples/tests/test_notebooks.py
@@ -9,10 +9,10 @@ from nbclient import NotebookClient
 NOTEBOOK_DIR = Path("notebooks/examples/")
 NOTEBOOK_FILES = list(NOTEBOOK_DIR.glob("*.ipynb"))
 ALLOWED_ERROR_NAMES = [
-    "requests.exceptions.ConnectionError",
-    "requests.exceptions.ConnectTimeout",
-    "requests.exceptions.HTTPError",
-    "requests.exceptions.ReadTimeout",
+    "ConnectionError",
+    "ConnectTimeout",
+    "HTTPError",
+    "ReadTimeout",
 ]
 
 
--- a/notebooks/examples/tests/test_notebooks.py
+++ b/notebooks/examples/tests/test_notebooks.py
@@ -13,6 +13,10 @@ ALLOWED_ERROR_NAMES = [
     "ConnectTimeout",
     "HTTPError",
     "ReadTimeout",
+    # NameError cascades from network failures: when a cell that makes a
+    # network call is allowed to fail, subsequent cells that reference its
+    # output variable will raise NameError.
+    "NameError",
 ]
 
 
--- a/src/dysh/line/tests/conftest.py
+++ b/src/dysh/line/tests/conftest.py
@@ -2,6 +2,9 @@ import functools
 
 import pytest
 import requests
+from astropy.table import Table
+
+import dysh.line.search as _search_module
 
 SPLATALOGUE_NETWORK_EXCEPTIONS = (
     requests.exceptions.HTTPError,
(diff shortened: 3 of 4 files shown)


mpound:
Here's the complete picture of what's failing and why (after the initial PR):

  ---
  Three separate failures
  
  1. Windows 3.10 — line_search.ipynb (NameError cascade)

  - Cell 6: Calls Splatalogue online — raises ConnectTimeout (which IS in ALLOWED_ERROR_NAMES, so
  nbclient continues)
  - Cell 7: print(f"{len(table)} rows returned.") — table was never assigned because cell 6 raised an
  exception before returning → NameError (NOT in ALLOWED_ERROR_NAMES) → test fails

  This is a direct consequence of the branch's fix: the ALLOWED_ERROR_NAMES fix works correctly for the
  network error, but the downstream cells that depend on the failed cell's output still crash with
  NameError.

  2. Windows 3.12 — hi_survey.ipynb (CellTimeoutError)

  - from_url("https://greenbankobservatory.org/~koneil/HIsurvey/results_all.dat", ...) in cell 114 uses
  httpx.Client with no connection timeout
  - On Windows CI runners, this connection appears to hang for > 600 seconds
  - The reported timed-out cell is cell 115 (the file-reading cell), which is somewhat confusing —
  likely cell 112 (a heavy loop over all sources) consumed most of the 600s budget, leaving cell 115 to
  trip the timer

  3. Windows 3.11 — uv install failure (transient infrastructure)

  - Failed fetching https://raw.githubusercontent.com/astral-sh/versions/main/v1/uv.ndjson — transient
  GitHub network hiccup, not a code issue.

  ---
  Fixes suggested:

  - For the NameError cascade (line_search.ipynb and potentially others): add "NameError" to
  ALLOWED_ERROR_NAMES in test_notebooks.py. This is pragmatic — the test framework already accepts that
  network-dependent cells may fail; cascading NameErrors in subsequent cells are a predictable side
  effect.

  - For hi_survey.ipynb timeout: add an HTTP timeout to from_url in src/dysh/util/download.py so it fails
  fast rather than hanging:
  client = httpx.Client(follow_redirects=True, timeout=30.0)
  This alone won't fix the test (cell 115 would then fail with NameError from table_file being
  undefined), but combined with NameError in the allowed list, it would pass.

astrofle:
~~The notebook execution can still fail because if a cell error is ignored because of Splatalogue being down, then any cells that require the output (see e.g., cell 4 in line_search.ipynb) will fail for another reason, like an undefined variable (see e.g., https://github.com/GreenBankObservatory/dysh/actions/runs/27228197344/job/80606431844?pr=1122). What a mess.~~
I see you already fixed this. Thanks.

mpound:
> ~The notebook execution can still fail because if a cell error is ignored because of Splatalogue being down, then any cells that require the output (see e.g., cell 4 in line_search.ipynb) will fail for another reason, like an undefined variable (see e.g., https://github.com/GreenBankObservatory/dysh/actions/runs/27228197344/job/80606431844?pr=1122). What a mess.~ I see you already fixed this. Thanks.

Looks like `hi_survey.ipynb` can still fail due to timeout, but not because of Splatalogue. 
https://github.com/GreenBankObservatory/dysh/actions/runs/27298647107/job/80638057705?pr=1122

src/dysh/util/download.py
```diff
@@ -34,7 +34,7 @@ def from_url(url, path=Path(".")):
 
     try:
         # Make the HTTPX client
-        client = httpx.Client(follow_redirects=True)
+        client = httpx.Client(follow_redirects=True, timeout=30.0)
```
astrofle: Is this still needed after [this CI failure](https://github.com/GreenBankObservatory/dysh/actions/runs/27298647107/job/80638057705?pr=1122)?
The commit says:
"Add 30s timeout to httpx.Client in from_url so GBO server hangs on Windows CI runners fail fast with ConnectTimeout instead of blocking a cell for 600+ seconds"
which seems to be the error of the latest CI failure. This suggests that adding ``timeout=30.0`` is not helping. Could this have unintended side effects? If so, I think we should remove this change.

mpound: Yes, this is still a failure on cell 67 ``table_file=from_url(...)``.   So it seems `timeout=30.0` is being ignored. I can remove it (or merge #1124 ) but the timeout bug for `from_url` will remain.  Arguably, that is not related to the Splatalogue fix, so should be dealt with in a different PR, allowing this fix to be merged.

mpound: I removed the timeout from `from_url`

mpound:
Possibly we could add ``CellTimeoutError`` to the list of ALLOWED_ERROR_NAMES.   This might fix the `from_url` problem.

astrofle (approved):
Thanks

PR #1123, opened 2026-06-10
Speedup RTD option B

astrofle:
Trying option B. Disbling RTD notebook execution and trying to use notebook output uploaded to GitHub.


tchamberlin:
Tweaked this so that local behavior doesn't change, but we still skip nb builds in RTD. Also added a commit that rebuilds all the notebooks and updates them in the repo

mpound:
Added an RTD build to track this:
https://app.readthedocs.org/projects/dysh/builds/?version__slug=speedup_rtd_b

astrofle:
> Tweaked this so that local behavior doesn't change, but we still skip nb builds in RTD. Also added a commit that rebuilds all the notebooks and updates them in the repo

An issue (due to the mess we have with the plotting), the plots are empty in the notebooks. I'm guessing that https://github.com/GreenBankObservatory/dysh/blob/df35016fd2c216a0c736edf47e38437ce7ab49d4/src/dysh/util/core.py#L583 is returning ``False`` when the notebooks are executed like you did.

tchamberlin:
Okay for reference I ran

```sh
$ READTHEDOCS=True MPLBACKEND=Agg ls notebooks/examples/*.ipynb \
  | xargs -P 4 -I{} uv run jupyter nbconvert --to notebook --execute --inplace {}
```

Not sure if we have a sanctioned way of doing this? Obviously the notebooks build as part of sphinx build but they output somewhere else

astrofle:
> Okay for reference I ran
> 
> ```shell
> $ READTHEDOCS=True MPLBACKEND=Agg ls notebooks/examples/*.ipynb \
>   | xargs -P 4 -I{} uv run jupyter nbconvert --to notebook --execute --inplace {}
> ```
> 
> Not sure if we have a sanctioned way of doing this? Obviously the notebooks build as part of sphinx build but they output somewhere else

I don't know why the above results in ``os.environ.get("READTHEDOCS")`` being ``None``.

I tried:
```shell
export READTHEDOCS=True; ls notebooks/examples/*.ipynb \
| xargs -P 4 -I{} uv run jupyter nbconvert --to notebook --execute --inplace {}
```
and it worked (the notebooks have visible figures).

mpound (approved):

PR #1130, opened 2026-06-22
Ensure lazy flag temporary files  are cleaned up as we go.

### LINKED ISSUE #1125: LazyFlagArray.to_dense() writes many GBs of tmp files (fills up /tmp on GBO hosts)
**Describe the bug**

On large datasets, standard dysh use yields massive amounts of `*.flags` files on disk in `/tmp`. These quickly fill up the small /tmp dirs on GBO data processing hosts, and lead to weird crashes in dysh and other programs

This appears to be a result of `LazyFlagArray.to_dense()` materializing the entire `(nrows, nchan)` bool mask, which, if >1GB, are backed by memmap into temp files. Every call results in a new temp file (no caching)

I think ideally we would be streaming data into the output FITS files directly. But a simpler option would be to just move the tmp dir location elsewhere. Perhaps respect `XDG_CACHE_HOME`? Then we could potentially override that in a GBO-specific way if needed 

PR #1103 does _not_ fix this; it is only a partial mitigation, because these files persist for the lifetime of the dysh process, and are written to `/tmp`

**How to Reproduce**
```py
import glob

# A session whose flag mask exceeds 1 GB
sdf = GBTFITSLoad("some_large_session.raw.vegas")

print(glob.glob("/tmp/*.flags"))   # ideally empty...

# Any all-rows operation that sets the mask will trigger call to to_dense()
sdf.getps(scan=...).timeaverage()  # read path
sdf.write("out.fits", flags=True, overwrite=True)  # write path

print(glob.glob("/tmp/*.flags"))   # multi-GB temp file(s), which persist until the process exits cleanly
```

**Environment**
- Dysh version: 1.1

mpound:
This fixes #1125 
1.  Add `weakref.finalize` in `LazyFlags.to_dense()` so that memory mapped files are removed when there are no longer strong references to the arrays they underlie.
2. Use $DYSH_SCRATCH as the temporary directory if it is set. 
3. Add tests that all this works as advertised.  Windows OS is skipped for one test because of delay in actually removing files on Windows.
4. Add a Note to the quick reference guide about the need for DYSH_SCRATCH (or TMPDIR)

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -41,7 +41,7 @@ We refer to this interface as the dysh shell.
 
 .. note::
 
-   For large SDFITS files (> 1 GB or so), dysh may create temporary files during certain operations to hold flag arrays.  By default these files are created in ``/tmp``. To avoid filling up ``/tmp``, you can (and should) change that default by setting ``$DYSH_SCRATCH`` or ``$TMPDIR`` environment variables to point to a directory where temporary files can be written.
+   For large SDFITS files (> 1 GB or so), dysh may create temporary files during certain operations to hold flag arrays.  By default these files are created in ``/tmp``. To avoid filling up ``/tmp``, you can (and should) change that default by setting ``$DYSH_SCRATCH`` or ``$TMPDIR`` environment variables to point to a directory where temporary files can be written.  Note these must be set before you start dysh as the values are cached by Python at startup, so ``os.putenv()`` would not have any effect..
 
 Loading Data
 ============
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -39,6 +39,10 @@ After being installed, the ``dysh`` command will be available through the comman
 This will launch an `iPython <https://ipython.readthedocs.io/en/stable/>`_ session with some modules and classes pre-loaded (e.g., `~dysh.fits.gbtfitsload.GBTFITSLoad`), and with logging.
 We refer to this interface as the dysh shell.
 
+.. note::
+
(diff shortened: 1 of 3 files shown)


docs/source/index.rst
```diff
@@ -39,6 +39,10 @@ After being installed, the ``dysh`` command will be available through the comman
 This will launch an `iPython <https://ipython.readthedocs.io/en/stable/>`_ session with some modules and classes pre-loaded (e.g., `~dysh.fits.gbtfitsload.GBTFITSLoad`), and with logging.
 We refer to this interface as the dysh shell.
 
+.. note::
+
+   For large SDFITS files (> 1 GB or so), dysh may create temporary files during certain operations to hold flag arrays.  By default these files are created in ``/tmp``. To avoid filling up ``/tmp``, you can (and should) change that default by setting ``$DYSH_SCRATCH`` or ``$TMPDIR`` environment variables to point to a directory where temporary files can be written.
```
astrofle: Setting TMPDIR seems to make no difference (the files are still being written to ``/tmp``). DYSH_SCRATCH works as advertised.

astrofle (changes requested):
All looks great, except the bit about using TMPDIR.
Thanks a lot!

docs/source/index.rst
mpound: TMPDIR is by default checked by Python tempfile: https://docs.python.org/3/library/tempfile.html .
Does `uv run pytest -k test_to_dense_uses_tmpdir_env` pass in your environment?  If yes, then TMPDIR is working as advertised.

astrofle: The test passes, however, the logger still says it is writting to ``/tmp``:
Screenshot 2026-06-23 at 2 08 17 PM

mpound:
Did you set TMPDIR before starting dysh? The environment is cached, so ``os.putenv(TMPDIR,"/foo/bar")`` will not be reflected in ``tempfile.gettempdir()``

astrofle (approved):
That explains it. Thanks

mpound:
> That explains it. Thanks

I'll add that to the documentation.

PR #1136, opened 2026-06-23
Issue 1119 changing spectral axis

### LINKED ISSUE #1119: switching to a spectrum with a km/s axis fails to set some metadata
**Describe the bug**
I converted a spectrum to one where the spectral axis is km/s, this was the way i found I could write a spectrum with a km/s column.  However this spectrum is missing lots of meta data, e.g. one cannot plot, or use cog().

Perhaps the question should be:   what is the correct way to switch between units on the spectral axis

My reason to fall into this problem is a more general usage of dysh:  being able to read an ascii spectrum (e.g. from NED), manually set the units (usually they are km/s in NED), maybe reference frame, although for some computations such as cog() these should not be important.

**How to Reproduce**

```
sp1 = Spectrum.fake_spectrum()
sp2 = sp1.with_spectral_axis_unit("km/s")
sp2.plot()    # produces a plot, but doesn't show meta data
sp2.cog()     # fails on a meta data 
```

**Environment**
- Dysh version  1.0
- Python version  3.12
- OS Ubuntu 26.04

astrofle:
Adds Spectrum set_spectral_axis and with_spectral_axis_unit (overloads the specutils method).
    
``set_spectral_axis`` modifies the units, frame or Doppler convention of the spectrum. 
``with_spectral_axis_unit`` adds the observer and target attributes back to the new spectrum object to make it compatible with ``dysh.spectra.Spectrum`` objects.

Example:
```Python
s1 = Spectrum.fake_spectrum()
s1.set_spectral_axis(unit="km/s", doppler_convention="radio") # Changes the Spectrum in place.
s2 = s1.with_spectral_axis_unit("km/s", doppler_convention="optical") # Creates a new Spectrum.
```


mpound (changes requested):
The example fails for me because ``specutil.Spectrum.with_spectral_axis_unit`` uses ``velocity_convention`` as the keyword not `doppler_convention` so passing ``*args`` raises an exception.
```
Cell In[6], line 4
      2 s1 = Spectrum.fake_spectrum()
      3 s1.set_spectral_axis(unit="km/s", doppler_convention="radio") # Changes the Spectrum in place.
----> 4 s2 = s1.with_spectral_axis_unit("km/s", doppler_convention="optical") # Creates a new Spectrum.

File /bigdisk/src/dysh/src/dysh/spectra/spectrum.py:1101, in Spectrum.with_spectral_axis_unit(self, *args, **kwargs)
   1100 def with_spectral_axis_unit(self, *args, **kwargs):
-> 1101     spec = super().with_spectral_axis_unit(*args, **kwargs)
   1102     spec._observer = self.observer.copy()
   1103     spec._spectral_axis.observer = self.spectral_axis.observer.copy()

TypeError: OneDSpectrumMixin.with_spectral_axis_unit() got an unexpected keyword argument 'doppler_convention'
```

specutils.__version is '2.3.0'

The tests pass, though.

src/dysh/spectra/spectrum.py
```diff
@@ -1097,6 +1097,73 @@ def rest_value(self, value: Quantity):
...
+
+            .. note: This must be the rest frequency/wavelength *in vacuum*,
+                     even if your spectrum has air wavelength units
+
+        """
+        # The docstring should be kept up to date.
+        # Keep docstring empty to re-use the one from specutils.
+        # Keeping the specutils doc introduces yet another name
+        # for what we call `doppler_convention`, `velocity_convention`.
+        # This is a source of frustration...
```
mpound, line 1131: Because specutils does not follow the astropy parameter name 'doppler_convention' in `SpectralQuantity`

notebooks/examples/velocity_frames.ipynb
mpound, line 1: Cell 6: sdfits["CTYPE1"][0]  returns nan

mpound: Also, sometimes notebook plots just hang but I don't belive that has anything to do with this PR.

mpound (approved):
Very good, thanks!

notebooks/examples/velocity_frames.ipynb
astrofle: Fixed the cell output being NaN.

PR #1137, opened 2026-07-08
Performance series PR 0: workflow benchmark harness and before/after protocol

mpound:
## Summary

First PR of the performance-enhancement series that recovers the gains from closed PR #1049 in small, reviewable chunks. The full multi-PR plan is in `perfenhance_plan.md` (added here). **No `src/` changes** — this PR establishes the measurement infrastructure that every subsequent PR in the series will use to report before/after numbers.

## What's included

- **`benchmark/workflows/`** — the dysh-vs-GBTIDL workflow harness from PR #1049 (`run_bench.py` plus `hi_survey`, `argus_vanecal`, `nod_kfpa`, and `exit` benchmark scripts), with three adaptations:
  - Data paths resolve via `--data-path` → `$DYSH_BENCH_DATA_PATH` → canonical GBO path → `dysh_data` alias, so the benchmarks run both at GBO and elsewhere.
  - New `--capture-golden` flag and `golden.txt` fallback: `--verify` compares dysh output against GBTIDL when `gbtidl` is on PATH, otherwise against a committed golden capture, so correctness gating works off-GBO.  The `golden` file is the reference baseline of known correct output for each benchmark.  The reason they exist at all is that the original PR #1049 harness verified dysh against a live GBTIDL run, which only works at GBO where gbtidl is installed. The golden files stand in for GBTIDL on machines without GBTIDL, so the correctness gate works everywhere. At GBO, `--verify`  still prefers the real GBTIDL comparison automatically.
  - Removed dead `hasattr` code paths for methods that only existed on the old PR #1049 branch.
- **Golden captures** for `hi_survey` (Session02) and `argus_vanecal` (TGBT22A_603_05), taken on this unmodified release-1.2 baseline. A performance PR must not change these values.
- **`benchmark/bench_calibration.py`** — DTime micro-driver isolating per-spectrum costs: first-vs-warm `getps`, `getspec` with/without WCS, `timeaverage`, and common Spectrum operations.
- **`benchmark/README.md`** — the before/after protocol (environment pinning, warm/cold cache rules, cProfile hotspot confirmation, regression gate) required in every performance PR description.
- **`perfenhance_plan.md`** — the full plan for the PR series.

## Baseline signals (local host, SSD, OMP_NUM_THREADS=1 [Marc's laptop])

- `getspec` with WCS: 1518 ms vs 217 ms without, for 11 spectra (per-spectrum WCS construction)
- First `getps`: 710 ms vs ~350 ms warm (lazy-load penalty)
- `hi_survey`: 16.2 s script time, ~70% in the four getsigref+timeaverage stages
- `argus_vanecal`: 111 s, of which 102 s is the 16-feed vanecal loop

## Test plan

- [x] All four workflows run locally end-to-end via `run_bench.py`
- [x] `--verify` passes for `hi_survey` and `argus_vanecal` against their golden captures (0.00% difference)
- [x] `bench_calibration.py` runs on the `getps` example dataset
- [x] `ruff check` / `ruff format --check` clean on all added files
- [x] No `src/` changes, so the unit test suite is unaffected

## Note for reviewers

The local argus data used for the golden capture is the full TGBT22A_603_05 session rather than GBO's vanecal-only subset, so the `GBTFITSLoad` stage is inflated in local timings; the vanecal stages and golden Tsys values are unaffected. At GBO, the canonical subset path and real GBTIDL comparison are used automatically.

🤖 Partially generated with [Claude Code](https://claude.com/claude-code)


mpound:
Eyeballs especially  needed on the overall plan for 9 PRs.

astrofle:
> Eyeballs especially needed on the overall plan for 9 PRs.

Honestly, I find it a bit hard to grasp all the details of the plan (I find it too technical for my sleep deprived self), but overall it looks reasonable; create PRs for the things that speed up existing portions of the code based on PR #1049.
It is unclear how a change will be judged. I'd guess: the results from the "golden" files must be reproduced, the execution time must be lower than the benchmark, the CI tests must pass. Any other metrics?

Do we need to merge this PR? (I hope not.)

mpound:
> Honestly, I find it a bit hard to grasp all the details of the plan (I find it too technical for my sleep deprived self), but overall it looks reasonable; create PRs for the things that speed up existing portions of the code based on PR #1049. It is unclear how a change will be judged. I'd guess: the results from the "golden" files must be reproduced, the execution time must be lower than the benchmark, the CI tests must pass. Any other metrics?

Yes, those metrics.  Execution time speedup must be significant (however we define that) to make the PR worth it.    Or, if speedup is marginal, code must be obviously cleaner to make it worth it.  

> Do we need to merge this PR? (I hope not.)

Yes (to release-1.2), because it contains the benchmark harness from PR #1049 (with only minor changes).  

@tchamberlin @teuben Can you review the plan [``benchmark/perfenhance_plan.md``](https://github.com/GreenBankObservatory/dysh/blob/perf-pr0-benchmark-harness/benchmark/perfenhance_plan.md) and see if it makes sense to you?

I don't intend to speed through these PRs. There is a lot in them even with the smaller chunks, so I want to make sure I (and we) really understand what they do.

perfenhance_plan.md
astrofle, line 1: This should be removed from here.

mpound: I put it in so reviewers could easily see and possibly amend it.   Will remove before merge.

astrofle:
I think we should add more things to be verified and be more strict about the tolerance. Currently, the verify scripts would only check system temperatures for Argus, and the blue and red wing fluxes for the HI survey. We need to add, at least:
* Spectral axis comparisons for averages and individual integrations
* System temperature for hi_survey
* Coordinates for averages (Spectrum)
* Metadata for averages (Spectrum)

Ideally, the above would also be propagated to the GBTIDL scripts, but given our resources, maybe we can live without it.

A 2% tolerance is larger than the current standard of mK with respect to GBTIDL (Argus system temperatures are ~100s so 2% is ~2 K, x1000 larger) -- I might be off since this is relative to the current values and not absolute.

tchamberlin:
Ya these sorts of enhancements are really only possible to the extent that we can fully validate the result, and I'm not familiar enough with the project to speak to how much 100% passing tests really means.

And certainly when I designed the benchmark harness itself there's a very good chance that I skipped some important bits due to same lack of understanding. So I think the starting point here would need to be a human-understandable harness, that we are confident represents both the GBTIDL and dysh sides accurately.

My goal was pretty narrow originally: see how much perf gain I could get while keeping the tests green. There wasn't much care given to the harness itself, other than "get something working". And definitely for the code itself; as you mention it's pretty opaque and would need to be human-understandable before merge.

Maybe an implementation PR based on this one might help demonstrate what that would look like? i.e. a fully human written PR, with de-slopped code that covers a single, easy(ish) to understand enhancement to the code?

mpound:
> I think we should add more things to be verified and be more strict about the tolerance. Currently, the verify scripts would only check system temperatures for Argus, and the blue and red wing fluxes for the HI survey. We need to add, at least:
> 
> * Spectral axis comparisons for averages and individual integrations

Are you saying here ``spectral_axis(dysh) == spectral_axis(gbtidl)`` or ``spectral_axis(dysh)_avg = spectral_axis(dysh)_i`` for each integration? 

> * System temperature for hi_survey

Can be added.

> * Coordinates for averages (Spectrum)
> * Metadata for averages (Spectrum)

Again is this averages vs gbtidl or vs individual dysh integrations?   If the latter, I think that regular pytests should deal with that.

> Ideally, the above would also be propagated to the GBTIDL scripts, but given our resources, maybe we can live without it.
> 
> A 2% tolerance is larger than the current standard of mK with respect to GBTIDL (Argus system temperatures are ~100s so 2% is ~2 K, x1000 larger) -- I might be off since this is relative to the current values and not absolute.

Agreed. I will add changing tolerance to 1 mK, compare Tsys, and (depending on your answer) average comparison as PR 0.5 in the plan.

astrofle:
> > I think we should add more things to be verified and be more strict about the tolerance. Currently, the verify scripts would only check system temperatures for Argus, and the blue and red wing fluxes for the HI survey. We need to add, at least:
> > 
> > * Spectral axis comparisons for averages and individual integrations
> 
> Are you saying here `spectral_axis(dysh) == spectral_axis(gbtidl)` or `spectral_axis(dysh)_avg = spectral_axis(dysh)_i` for each integration?
> 
> > * System temperature for hi_survey
> 
> Can be added.
> 
> > * Coordinates for averages (Spectrum)
> > * Metadata for averages (Spectrum)
> 
> Again is this averages vs gbtidl or vs individual dysh integrations? If the latter, I think that regular pytests should deal with that.
> 
> > Ideally, the above would also be propagated to the GBTIDL scripts, but given our resources, maybe we can live without it.
> > A 2% tolerance is larger than the current standard of mK with respect to GBTIDL (Argus system temperatures are ~100s so 2% is ~2 K, x1000 larger) -- I might be off since this is relative to the current values and not absolute.
> 
> Agreed. I will add changing tolerance to 1 mK, compare Tsys, and (depending on your answer) average comparison as PR 0.5 in the plan.

That's a better idea. We need to have more verification tests for spectral axes, metadata, etc... Currently we are blind to changes to those things that could have a real scientific impact.

benchmark/workflows/scripts/argus_vanecal/dysh_script.py
teuben, line 1: where the code says  `if twarm >= 370.0:` there should be a comment this was a faulty sensor (or whichever) that happened mostly during early 2025 (?) bla  bla.   I'd prefer we have have this type of code centralized so it can be re-used. I remember hacking this twice in the degas and gbtpipe code.

teuben, line 1: this code introduced a `DYSH_BENCH_DATA_PATH` environment variable.  Is that documented somewhere.  I know we have `DYSH_DATA` and `SDFITS_DATA`, which are in the glossary.   Althiough this one is a developer setting, still useful to properly advertise it.

perfenhance_plan.md
teuben, line 1: this file appears twice:
```
./benchmark/perfenhance_plan.md
./perfenhance_plan.md
```
as seen from the dysh root

teuben (changes requested):

benchmark/workflows/scripts/argus_vanecal/dysh_script.py
mpound: It is described in benchmark/workflows/README.md.   The README in benchmark is from our Q8-9 investigation, possibly should be renamed or content clarified.

mpound: I am not cognizant about this sensor (or whichever).  Can you provide more context?

teuben (approved):

astrofle (changes requested):
Could you change the base branch?
I don't think this should go into a release branch  or main, as it is an experiment that might not produce improvements (I've tried bits and pieces from the original PR and saw no improvements).

astrofle (approved):

astrofle:
I still think this should not go into a release branch or main, but won't block it.

mpound:
We've decided not to merge this, but rather wait until PR 0.5 when the benchmark code has been updated and possibly refactored with the existing benchmark code.

PR #1138, opened 2026-07-10
perf: accelerate data selection with duckdb and cache WCS/targets in getspec

danjsiegel:
## Overview

This PR introduces two targeted performance optimizations to the core calibration pipeline, addressing bottlenecks in both DataFrame indexing and coordinate/WCS construction. 

1. **DuckDB Fast-Path**: Offloads complex, multi-rule data selection and flag elimination from Pandas to DuckDB (as an optional dependency).
2. **WCS / Target Caching**: Prevents redundant `astropy` coordinate reconstructions across integrations within the same scan.

## 1. DuckDB Data Selection (`gbtfitsload.py`)

Previously, `_common_selection` executed a sequential `pandas.merge(how='inner')` for *each* active selection rule, followed by a Python set-difference `.isin()` check to drop flagged rows. This scales poorly as dataset size and rule counts grow.

This introduces an optional `duckdb` fast-path (installable via `dysh[fast]`) that collapses all rules and flag eliminations into a single vectorized C++ scan and `NOT EXISTS` anti-join. 

**Overhead Protection**: The DuckDB path is explicitly gated behind `len(self._selection._selection_rules) > 0`. If no selection rules are active, it falls back to the native pandas subsetting, ensuring that you never pay DuckDB query compilation overhead when there's no merge work to do.

### Scaling Benchmark

*Testing `getps` on a standard ACS dataset. Each rule forces an additional `pandas.merge` pass, whereas DuckDB evaluates the entire filter in one pass.*

| Rules | Pandas (ms) | DuckDB (ms) | Speedup |
|-------|-------------|-------------|---------|
| 0     | 75.3        | 75.5        | 1.00x *(DuckDB bypassed)* |
| 1     | 84.3        | 80.1        | 1.05x |
| 3     | 105.8       | 90.8        | 1.17x |
| 5     | 133.8       | 107.4       | 1.25x |
| 10    | 231.2       | 174.2       | **1.33x** |

**Takeaway:** Pandas grows linearly $\mathcal{O}(N)$ with the number of selection rules. DuckDB is sub-linear. 

## 2. WCS and Target Caching (`scan.py`, `spectrum.py`)

Profiling `getspec_wcs` revealed significant per-spectrum overhead. For a tracked source, the WCS header and celestial source position are identical across all integrations in a scan only `DATE-OBS` changes. Yet, `make_spectrum` was building a new `astropy.wcs.WCS` and `astropy.coordinates.SkyCoord` from scratch for every integration, triggering expensive ERFA validations each time.

**The Fix:**
- Updated `Spectrum.make_spectrum` to accept `wcs` and `target` as optional kwargs.
- Implemented lazy evaluation and caching in `ScanBase.getspec`. The `WCS` and `target` are built on the first integration (`i=0`) and reused for all subsequent integrations in the scan.

### `getspec_wcs` Benchmark

*Measuring warm execution of an 11-integration `getps` scan.*

| Path | Warm Execution Time (11 ints) |
|---|---|
| `getspec_nowcs` | ~60 ms |
| `getspec_wcs` (Before Cache) | ~230 ms |
| `getspec_wcs` (After Cache) | **~190 ms** |

**Takeaway:** This simple cache saves ~40 ms per 11-integration scan. 

*(Note: The remaining ~130 ms overhead for `getspec_wcs` is entirely isolated to `specutils.Spectrum1D.__init__` eagerly evaluating the `spectral_axis` across all ~8192 channels via `wcs.all_pix2world`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,10 +49,10 @@ nb = [
   "jupyter-app-launcher>=0.3.1",
   "jupyterlab-lsp>=5.2.0",
 ]
-#fast = [
-#  "rsdfits>=0.1.0",
-#]
-all = ["dysh[nb]"]
+fast = [
+  "duckdb>=0.10",
+]
+all = ["dysh[nb,fast]"]
 
 #[tool.uv.sources]
 #rsdfits = { path = "../rsdfits", editable = true }
--- a/src/dysh/fits/gbtfitsload.py
+++ b/src/dysh/fits/gbtfitsload.py
@@ -71,6 +71,7 @@ except ImportError:
 
 try:
     import duckdb
+
     _DUCKDB_AVAILABLE = True
 except ImportError:
     _DUCKDB_AVAILABLE = False
@@ -1949,9 +1950,7 @@ class GBTFITSLoad(SDFITSLoad, HistoricalBase):
         """
         for _name, _val in (("ifnum", ifnum), ("plnum", plnum), ("fdnum", fdnum)):
             if _val is not None and not isinstance(_val, (int, float, np.integer, np.floating)):
-                logger.debug(
-                    f"_common_selection_duckdb: {_name}={_val!r} is not a scalar, falling back to pandas path"
-                )
+                logger.debug(f"_common_selection_duckdb: {_name}={_val!r} is not a scalar, falling back to pandas path")
                 return None
 
         where_parts = []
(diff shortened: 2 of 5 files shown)


mpound:
@danjsiegel Thanks for these. I am working on a series of performance enhancements (PR #1137),  so I will review this PR during that process.  My list also includes WCS caching so it will be interesting to see if we both solved the same way.

danjsiegel:
> @danjsiegel Thanks for these. I am working on a series of performance enhancements (PR #1137), so I will review this PR during that process. My list also includes WCS caching so it will be interesting to see if we both solved the same way.

@mpound I went and looked at [#1137](https://github.com/GreenBankObservatory/dysh/pull/1137) and subsequently [#1049](https://github.com/GreenBankObservatory/dysh/pull/1049). They're actually looking mostly complementary, and solving slightly different bottlenecks. Your WCS work caches the template at the module level, while my getspec cache lives at the scan level (avoiding reconstruction entirely after the first integration). They actually look like they'd layer together nicely. Thanks for checking it out, happy to rebase whenever #1137 is merged

astrofle (commented):
Thanks a lot for the PR @danjsiegel 
I would reject it unless @mpound or @etsmit have other opinions/comments.

[As we discussed in an earlier attempt to cache WCS/targets](https://github.com/GreenBankObservatory/dysh/pull/1049#discussion_r2944056970), this has the drawback that reduces the scientific accuracy of the results as it does not properly capture changes in time and target. The effect is more pronounced for mapping scans, where both target and observing time change as a function of integration.

As for the duckdb changes, I am not familiar with the library, and feel like adding yet another split (_DUCKDB_AVAILABLE) in the code makes the code harder to maintain.

I appreciate pointing out potential speed ups and things that could be improved.

danjsiegel:
@astrofle no worries, scientific accuracy absolutely comes first. Completely understand keeping the dependency footprint clean without the optional DuckDB path as well. For whenever you guys tackle the larger performance roadmap, hopefully that helps that the bottleneck coming from `specutils.Spectrum1D.__init__`.

I'll go ahead and close this out. Thanks for taking the time to look through it!

PR #1143, opened 2026-07-29
Adds basic test for spectral axis

astrofle:
Test covers ScanBase.getspec

### DIFF
--- a/src/dysh/spectra/tests/test_scan.py
+++ b/src/dysh/spectra/tests/test_scan.py
@@ -554,6 +554,36 @@ class TestScanBase:
         # Channel selection in dysh is inclusive of the upper edge.
         assert np.all(tp.weights[channel[1] + 1 :] == pytest.approx(tp_sb[0].tsys_weight.sum()))
 
+    def test_getspec_spectral_axis(self):
+        """
+        Test to make sure the spectral axis calculation retains its scientific accuracy.
+        """
+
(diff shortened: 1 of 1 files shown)


src/dysh/spectra/tests/test_scan.py
```diff
@@ -554,6 +554,24 @@ def test_timeaverage_flags(self):
...
+        tp = sdf.gettp(scan=6, ifnum=0, plnum=0, fdnum=0)
+        s0 = tp[0].getspec(0)
+        sl = tp[0].getspec(2)
+
+        assert s0.meta["VELDEF"] == "RADI-LSR"
+        assert sl.meta["VELDEF"] == "RADI-LSR"
+
+        assert s0.spectral_axis[int(s0.meta["CRPIX1"]) - 1].value == s0.meta["CRVAL1"]
+        assert sl.spectral_axis[int(sl.meta["CRPIX1"]) - 1].value == sl.meta["CRVAL1"]
+
```
mpound, line 571: Don't we want to comparisons to `tp.timeaverage()`  also? (based on your comment in PR#1137).   Also check CDELT1 against spectral_axis channel width?

mpound (changes requested):
Good start.  I think the test needs a bit more in it.

src/dysh/spectra/tests/test_scan.py
astrofle: I added a test for CDELT1 and time averaging.

mpound (approved):
Good job, thanks!

Issue #1140, opened 2026-07-23
Invalid RADESYS column values


astrofle:
**Describe the bug**
The resulting SDFITS can have the RADESYS column set to something like "galactic", which is not a valid value for this column in the WCS standard. (this change was made while patching the creation of Spectrum objects that did not have a RADESYS keyword.) This invalid column value propagates downstream and ends up causing things like FITS cubes with invalid coordinates.

**How to Reproduce**
WIP

**Environment**
- dysh version 1.1.0

PR #1144, opened 2026-07-29
Fix for issue #1140

astrofle:
Directly parses the CTYPE2 and CTYPE3 columns to generate a valid celestial frame rather than updating the RADESYS column values.

### DIFF
--- a/docs/source/whatsnew/CHANGES.rst
+++ b/docs/source/whatsnew/CHANGES.rst
@@ -2,6 +2,13 @@
 What's New?
 ***********
 
+Version 1.2.1
+=============
+
+.. _v1-2-1-bugfixes:
+
+- Avoids updating the RADESYS column (`#1140 <https://github.com/GreenBankObservatory/dysh/issues/1140>`_)
+
 Version 1.2.0
 =============
 
--- a/src/dysh/__init__.py
+++ b/src/dysh/__init__.py
@@ -7,7 +7,7 @@ warnings.filterwarnings("ignore", message="invalid value encountered in do_forma
 # ipympl multiple-inheritance MRO issue with traitlets (ipympl#488)
 warnings.filterwarnings("ignore", message="Passing unrecognized arguments to super")
 
-__version__ = "1.2.0"
+__version__ = "1.2.1"
 
 __all__ = ["version", "system_info"]
 
--- a/src/dysh/coordinates/core.py
+++ b/src/dysh/coordinates/core.py
@@ -516,6 +516,23 @@ def change_veldef(ctype, toframe):
     return newctype
 
 
+def celestial_ctype_to_frame(ctype2, ctype3, radesys):
+    """ """
+
+    if ctype2 == "AZ" and ctype3 == "EL":
+        frame = "altaz"
+    elif ctype2 == "HA":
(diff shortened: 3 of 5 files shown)


mpound (changes requested):
The code looks fine, but it is hard for me to evaluate workings or its side effects without a minimally failing example.   Seems to work ok when using an index file.  Are there cases when RADESYS is needed but unset?

mpound:
Also it is unclear to me whether we should be pulling against main now or release-1.2

astrofle:
> The code looks fine, but it is hard for me to evaluate workings or its side effects without a minimally failing example. Seems to work ok when using an index file. Are there cases when RADESYS is needed but unset?

No. I patched the cases were RADESYS was unset to something that I thought was valid, based on how ``Spectrum`` was being constructed. That has now been removed because it was wrong. RADESYS is a standard FITS WCS keyword, with only five allowed values ([Calabretta & Greisen 2002](https://ui.adsabs.harvard.edu/abs/2002A%26A...395.1077C/abstract)). We should not touch that. If it is not set or wrongly set, it should be handled upstream (i.e., sdfits filler).

Here's a minimum failing example. The example requires the ``gbtgridder``, and [SAOImageDS9](https://sites.google.com/cfa.harvard.edu/saoimageds9) (I have not found another WCS library that would complain):

```Python
import numpy as np
from dysh.util.files import dysh_data

filename = dysh_data(example='otf1')

sdf = GBTFITSLoad(filename)

# Reproduce the wrong behavior.
# Since this data was observed in J2000, 
# we force it to look like Galactic coordinates 
# and set radesys to what the "old" code would have done.
# Force CTYPE2 and 3 to be Galactic coordinates, 
sdf["CTYPE2"] = "GLON"
sdf["CTYPE3"] = "GLAT"
# and then make RADESYS "galactic".
sdf["RADESYS"] = "galactic"

# Write it with the invalid RADESYS value.
# Only mapping scans to keep it lean.
sdf.write("test_radesys.fits", scan=list(np.r_[14:28:1]), overwrite=True)
```

Next steps require the ``gbtgridder`` and ds9. In a shell:
```bash
# Install the gbtgridder:
pip install "gbtgridder @ git+https://github.com/GreenBankObservatory/gbtgridder@release_3.0"

# Grid the data into a cube:
gbtgridder test_radesys.fits -o test_radesys --autoConfirm
```

Open the resulting ``test_radesys_cube.fits`` with ds9 (ds9 test_radesys_cube.fits). It will print this error:
```bash
!! AST: Error in routine fits2ast at line 3509 in file frame/fitsimage.C.
!  astRead(FitsChan): FITS keyword 'RADESYS' has the unrecognised value 'galactic'.
```
and the WCS coordinates in ds9 will not be available, only pixel coordinates.

astrofle:
> Also it is unclear to me whether we should be pulling against main now or release-1.2

I thought we were going back to how we were doing things before the review: any changes go into main, and then we "freeze" when we are "ready" for a release.

etsmit (approved):

mpound:
Ok I wasnt sure if the GBO installation was tracking main which could leave
it unstable.

mpound (approved):

PR #1146, opened 2026-08-10
Fix for issue #1126: hour angle in plot header

### LINKED ISSUE #1126: Plot header produces wrong hour angle
**Describe the bug**
The hour angle reported in a plot header can be wrong when the coordinates are not RADec, i.e. CRVAL2 is not RA.

**How to Reproduce**
```Python
from dysh.fits import GBTFITSLoad

fnm = get_project_testdata() / "AGBT18B_354_03/AGBT18B_354_03.raw.vegas"
sdf = GBTFITSLoad(fnm)
# Set hour angle and declination coordinates.
sdf["RADESYS"] = "hadec"
sdf["CTYPE2"] = "HA"

tp = sdf.gettp(scan=6, plnum=0, ifnum=0, fdnum=0)[0].total_power(0)

tp.target.ha.value # -4.776572061733339

# Now the dysh calculation.
from dysh.coordinates import ra2ha
ra2ha(tp.meta["LST"], 288.3514190739999) # -2.03
```

I can reproduce the dysh result using:
```Python
from astropy.time import Time
from astropy.coordinates import Angle
import astropy.units as u

t = Time(tp.meta["DATE-OBS"], scale='utc', location=tp.target.location)
r = Angle(tp.meta["CRVAL2"]*u.deg)
(t.sidereal_time('mean') - r).value # -2.0346590144912327
```

So, the problem is that the ``dysh`` calculation assumes that CRVAL2 is RA, when it might not be.

**Environment**
- Dysh version 1.0.1
- Python version
- OS

astrofle:
Instead of using a custom function to compute the hour angle, leverage the coordinate transformations of SkyCoord objects. Fix sanitize coordinate to preserve location and time information to enable coordinate transformation that require it.

### DIFF
--- a/docs/source/whatsnew/CHANGES.rst
+++ b/docs/source/whatsnew/CHANGES.rst
@@ -8,6 +8,7 @@ Version 1.2.1
 .. _v1-2-1-bugfixes:
 
 - Avoids updating the RADESYS column (`#1140 <https://github.com/GreenBankObservatory/dysh/issues/1140>`_)
+- Prints the correct hour angle in plots when the coordinates are not Equatorial (`#1126 <https://github.com/GreenBankObservatory/dysh/issues/1126>`_)
 
 Version 1.2.0
 =============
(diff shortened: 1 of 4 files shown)


mpound (approved):
Code looks good. I verified the failing example in #1126 now works.  Thanks!

PR #1148, opened 2026-08-12
Add cap to astropy version

astrofle:
dysh does not work with astropy 8.0.0.

We would need to publish a release with this cap to keep users from running into issues (?).


etsmit:
What specifically breaks with Astropy 8?

astrofle:
> What specifically breaks with Astropy 8?

See: https://github.com/GreenBankObservatory/dysh/issues/1149

The config file issue can actually be a show stopper, as dysh will not start if it cannot create the config file (there's an [option to override this, --skip-config](https://dysh.readthedocs.io/en/latest/users_guide/launching.html), but it might not be obvious).

etsmit (approved):

PR #1163, opened 2026-09-18
Fix for issue #1162

### LINKED ISSUE #1162: getfs(channel=) fails if there is only one CAL state
**Describe the bug**
If there is only on CAL state in a data file, an exception is raised if one tries to select channels during calibration via getfs.

The specific line tries to compute `nchan` from `self._sigcalon`, which is empty in this case.   A more robust method of computing `nchan` should be used.   Note this may affect other calibration routines and just not been uncovered there yet.  And not just for `nchan` -- any code that tries to access an empty array due to missing CAL state would hit this. 

**How to Reproduce**

```
# Works fine because CAL has both states
fn=dysh_data(test='getfs')
sdfok=GBTFITSLoad(fn)
set(sdfok['CAL'])
sbok=sdfok.getfs(scan=20,ifnum=0,plnum=0,fdnum=0)
sbok=sdfok.getfs(scan=20,ifnum=0,plnum=0,fdnum=0,channel=[1024,8192])

# Fails because CAL has only one state
sdf=GBTOffline("AGBT25B_386_01")
set(sdfok['CAL'])
sb2=sdf.getfs(scan=30,ifnum=0,plnum=0,fdnum=8,chan=[1024,8192])


File /bigdisk/src/dysh/src/dysh/spectra/scan.py:2780, in FSScan.__init__(self, gbtfits, scan, sigrows, calrows, fdnum, ifnum, plnum, bintable, calibrate, fold, shift_method, use_sig, smoothref, apply_flags, tscale, ap_eff, surface_error, zenith_opacity, observer_location, tsys, tcal, nocal, channel, vane)
   2778     self._nchan = nchan
   2779 else:
-> 2780     self._nchan = len(self._sigcalon[0])
   2782 self._finish_initialization(
   2783     calibrate,
   2784     {"fold": fold, "shift_method": shift_method},
   (...)   2789     tcal=tcal,
   2790 )

IndexError: index 0 is out of bounds for axis 0 with size 0

```

**Environment**
Please provide the outputs of ``dysh.system_info()`` here (only available after dysh 1.1.0).
To get the outputs use: ``dysh -s``.

platform
--------
platform.platform() = 'Linux-4.18.0-553.158.1.el8_10.x86_64-x86_64-with-glibc2.28'
platform.version() = '# 1 SMP Sun Aug 23 20:58:55 EDT 2026'
platform.python_version() = '3.11.13'

packages
--------
dysh                 1.0.1
astropy              7.2.0
numpy                2.4.3
scipy                1.17.1
matplotlib           3.10.8
pandas               3.0.1
specutils            2.3.0
fitsio               1.3.0

astrofle:
Use a spectrum with the noise diode off to determine the number of channels.


mpound (approved):
If the requested number of channels is not even for `getfs`, you get a `ValueError: operands could not be broadcast together` from astropy quantity.py (which is really probably from numpy).  We should catch this in `getfs`.   Or better add a boolean option `force_even`to `util.get_valid_channel_range`   which defaults to False but is set to True in `getfs.`    This could be added to this PR or filed as a separate issue.

mpound:
> If the requested number of channels is not even for `getfs`, you get a `ValueError: operands could not be broadcast together` from astropy quantity.py (which is really probably from numpy). We should catch this in `getfs`. Or better add a boolean option `force_even`to `util.get_valid_channel_range` which defaults to False but is set to True in `getfs.` This could be added to this PR or filed as a separate issue.

Well I guess the option to switch `shift_method` as described in the docs covers this.   But its kind of buried so we need a better error message at least.

